PR #337, opened 2024-03-21
Print results table at the end of an evaluation session

### LINKED ISSUE #336: Print results at the end of an evaluation session

nkaenzig:
Closes #336 

image

### DIFF
--- a/configs/vision/dino_vit/offline/bach.yaml
+++ b/configs/vision/dino_vit/offline/bach.yaml
@@ -2,9 +2,12 @@
 trainer:
   class_path: eva.Trainer
   init_args:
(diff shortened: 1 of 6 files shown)


ioangatop:
@nkaenzig  could we make the matrix 3 columns, like mean, stdev and all (all the values separated by commas)? because in different smaller terminal this will not wrap nicely i think.. what do you think?

Also maybe we can round up to the 3d third digit, as the analytical results could be found in the output files

nkaenzig:
@ioangatop Good point, I updated the tables now accordingly (see screenshot above).

ioangatop (approved):

PR #351, opened 2024-04-03
Add `MultiEmbeddingsClassificationDataset`

### LINKED ISSUE #350: Add `MultiEmbeddingsDataset` for slide level tasks

nkaenzig:
Closes #350 

- Refactors embedding dataset classes
- Add a new `MultiEmbeddingsClassificationDataset` to support slide level tasks
  - The class is written in a way that is agnostic to WSIs
- The transformations `SampleFromAxis` and `Pad2DTensor` are included in this PR, because they are needed for common MIL setups in slide level tasks (e.g. our abmil implementation expects padded tensors of shape `(batch_size, n_instances, input_size)`)

### DIFF
--- a/src/eva/core/data/datasets/embeddings/base.py
+++ b/src/eva/core/data/datasets/embeddings/base.py
@@ -56,7 +56,7 @@ class EmbeddingsDataset(base.Dataset):
         self._root = root
         self._manifest_file = manifest_file
         self._split = split
-        self._column_mapping = column_mapping
+        self._column_mapping = default_column_mapping | column_mapping
         self._embeddings_transforms = embeddings_transforms
         self._target_transforms = target_transforms
 
--- a/src/eva/core/data/datasets/embeddings/base.py
+++ b/src/eva/core/data/datasets/embeddings/base.py
@@ -16,13 +16,13 @@ default_column_mapping: Dict[str, str] = {
     "path": "embeddings",
     "target": "target",
     "split": "split",
-    "slide_id": "slide_id",
+    "multi_id": "slide_id",
 }
 """The default column mapping of the variables to the manifest columns."""
 
 
 class EmbeddingsDataset(base.Dataset):
-    """Embeddings classification dataset."""
+    """Abstract base class for embedding datasets."""
 
     def __init__(
         self,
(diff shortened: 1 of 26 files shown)


src/eva/core/data/datasets/embeddings/base.py
```diff
@@ -54,12 +56,38 @@ def __init__(
         self._root = root
         self._manifest_file = manifest_file
         self._split = split
-        self._column_mapping = self.default_column_mapping | column_mapping
+        self._column_mapping = column_mapping
```
ioangatop, line 59: hmmm why do we change this? here the `self.default_column_mapping | column_mapping`, allows the user can to alter only one and the rest to remain the same (so he dont have to repeat the fields if they are the same as defaults), so now I think we have to repeat all of them

ioangatop (commented):
Looks great @nkaenzig! I'm taking a closer look, after we can merge

src/eva/core/data/datasets/embeddings/base.py
nkaenzig: good catch! I reverted this change now.

src/eva/core/data/datasets/embeddings/classification/multi_embeddings.py
```diff
@@ -0,0 +1,106 @@
...
+import torch
+from typing_extensions import override
+
+from eva.core.data.datasets.embeddings import base
+
+
+class MultiEmbeddingsClassificationDataset(base.EmbeddingsDataset):
+    """Dataset class for where a sample corresponds to multiple embeddings.
+
+    Example Usecase: Slide level dataset where each slide has multiple patch embeddings.
```
roman807, line 16: Usecase -> use case

src/eva/core/data/datasets/embeddings/classification/multi_embeddings.py
```diff
@@ -0,0 +1,106 @@
...
+        split: Literal["train", "val", "test"],
+        column_mapping: Dict[str, str] = base.default_column_mapping,
+        embeddings_transforms: Callable | None = None,
+        target_transforms: Callable | None = None,
+    ):
+        """Initialize dataset.
+
+        Expects a manifest file listing the paths of `.pt` files containing tensor embeddings.
+
+        The manifest is required to have a `column_mapping["multi_id"]` column that contains the
```
roman807, line 32: "The manifest is required to" --> "The manifest file must"

nkaenzig: Done

src/eva/core/data/datasets/embeddings/classification/multi_embeddings.py
nkaenzig: Done

ioangatop (approved):

PR #352, opened 2024-04-04
Disable Loggers that don't support saving to remote storage at runtime

### LINKED ISSUE #338: Support for saving results to remote storage
Officialy lightning supports remote storage:
https://lightning.ai/docs/pytorch/stable/common/remote_fs.html

Challenges:
- `TensorBoardLogger` and `CSVLogger` will raise errors when using azure blob storage.
- Disabling these loggers when setting the output directory to remote storage doesn't work because  `lightning.pytorch.callbacks.ModelCheckpoint` currently infers the directory to store checkpoints from the logger classes

nkaenzig:
Closes #338 

If `OUTPUT_ROOT` is set to a remote path and the config file lists loggers that don't support remote storage, the loggers will be disabled at runtime and a warning will be displayed:

image

With that change, we can now store the evaluation results directly to remote storage.

### DIFF
--- a/docs/reference/core/data/datasets.md
+++ b/docs/reference/core/data/datasets.md
@@ -2,7 +2,7 @@
 
 Reference information for the `Dataset` base class.
 
-::: eva.data.Dataset
+::: eva.core.data.Dataset
 
 ## Embeddings datasets
 ::: eva.core.data.datasets.EmbeddingsClassificationDataset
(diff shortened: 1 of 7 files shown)


ioangatop (commented):
actually cool trick!

ioangatop (approved):

PR #377, opened 2024-04-18
Add support for grid sampling to `WsiDataset`

### LINKED ISSUE #374: Add support for grid sampling to wsi dataset
Sampling method should be configurable, e.g. random, grid

nkaenzig:
Closes #374


src/eva/vision/data/wsi/patching/samplers.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class Sampler(abc.ABC):
+    """Base class for samplers."""
+
+    @abc.abstractmethod
+    def sample(
+        self,
+        height: int,
+        width: int,
```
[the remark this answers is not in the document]
nkaenzig: good catch, fixed now

src/eva/vision/data/datasets/wsi.py
```diff
@@ -55,49 +60,34 @@ def _wsi(self) -> wsi.Wsi:
...
+            sampler=self._sampler,
+        )
 
+    @override
+    def __getitem__(self, index: int) -> torch.Tensor:
+        x, y = self._coords.x_y[index]
+        width, height, level_idx = self._coords.width, self._coords.height, self._coords.level_idx
         patch = self._wsi.read_region((x, y), (width, height), level_idx)
-        patch = self._apply_transforms(patch)
+        patch = self._apply_transforms(torch.from_numpy(patch))
```
roman807, line 79: this seems to break the pipeline, since `torchvision.transforms.v2.ToImage()` permutes np.arrays but not tensors (see https://github.com/pytorch/vision/blob/main/torchvision/transforms/v2/functional/_type_conversion.py)

fixed in https://github.com/kaiko-ai/eva/pull/378

nkaenzig: Fixed, now. I ended up applying the dimension permute within the dataset class, to avoid typing issues with `_apply_transforms`.

src/eva/core/callbacks/writers/embeddings.py
```diff
@@ -90,8 +90,14 @@ def write_on_batch_end(
             embeddings_buffer, target_buffer = io.BytesIO(), io.BytesIO()
```
roman807, line 90: in line 89 we should replace `local_idx` with `global_idx` to prevent duplicate filenames (which could occur if batch_size < n_samples)

roman807 (approved):

PR #380, opened 2024-04-22
Reformat image segmentation datasets

### LINKED ISSUE #379: Fixes on segmentation base class and TotalSegmentator2D dataset
We will need to standardise the image and mask type to be compatible with torchvision v2 transforms standards, in order to use transformations easily

ioangatop:
Closes #379


src/eva/vision/data/datasets/segmentation/total_segmentator.py
```diff
@@ -134,19 +123,21 @@ def __len__(self) -> int:
         return len(self._indices) * self._n_slices_per_image
 
     @override
-    def load_image(self, index: int) -> np.ndarray:
+    def load_image(self, index: int) -> tv_tensors.Image:
         image_path = self._get_image_path(index)
         slice_index = self._get_sample_slice_index(index)
         image_array = io.read_nifti_slice(image_path, slice_index)
-        return image_array.repeat(3, axis=2)
+        image_rgb_array = image_array.repeat(3, axis=2)
```
nkaenzig, line 130: what is the `.repeat(3` here for?

src/eva/vision/utils/io/nifti.py
```diff
@@ -17,7 +17,7 @@ def read_nifti_slice(path: str, slice_index: int) -> npt.NDArray[Any]:
             return the full 3D image.
```
nkaenzig, line 17: `If `None`, it will return the full 3D image.`, is this still the case?

src/eva/vision/data/datasets/segmentation/total_segmentator.py
```diff
@@ -134,19 +123,21 @@ def __len__(self) -> int:
...
-    def load_mask(self, index: int) -> np.ndarray:
+    def load_masks(self, index: int) -> tv_tensors.Mask:
         masks_dir = self._get_masks_dir(index)
         slice_index = self._get_sample_slice_index(index)
         mask_paths = (os.path.join(masks_dir, label + ".nii.gz") for label in self.classes)
-        masks = [io.read_nifti_slice(path, slice_index) for path in mask_paths]
-        return np.concatenate(masks, axis=-1)
+        list_of_mask_arrays = [io.read_nifti_slice(path, slice_index) for path in mask_paths]
+        masks = np.concatenate(list_of_mask_arrays, axis=-1)
+        return tv_tensors.Mask(masks.transpose(2, 0, 1))
```
nkaenzig, line 140: I'm not sure if I understand this part 100%. Here we first concatenate on `axis=-1` (this corresponds to the image channel right?) and then we transpose from channel last to channel first format?

src/eva/vision/data/datasets/segmentation/total_segmentator.py
```diff
@@ -134,19 +123,21 @@ def __len__(self) -> int:
...
 
     @override
-    def load_mask(self, index: int) -> np.ndarray:
+    def load_masks(self, index: int) -> tv_tensors.Mask:
         masks_dir = self._get_masks_dir(index)
         slice_index = self._get_sample_slice_index(index)
         mask_paths = (os.path.join(masks_dir, label + ".nii.gz") for label in self.classes)
-        masks = [io.read_nifti_slice(path, slice_index) for path in mask_paths]
-        return np.concatenate(masks, axis=-1)
+        list_of_mask_arrays = [io.read_nifti_slice(path, slice_index) for path in mask_paths]
```
nkaenzig, line 138: How much overhead is there in opening the nifti file in each dataset item access?

nkaenzig (commented):
Looks nice. Left some minor comments

src/eva/vision/data/datasets/segmentation/total_segmentator.py
ioangatop: its because is a grey image and basically make it 3-channeled; maybe its better to use cv2 function though

src/eva/vision/utils/io/nifti.py
ioangatop: yes

src/eva/vision/data/datasets/segmentation/total_segmentator.py
ioangatop: exactly yes, because the torchvision Mask requires to have channel first

src/eva/vision/data/datasets/segmentation/total_segmentator.py
ioangatop: its not a lot, its like reading numpy files (like reading images)

src/eva/vision/data/datasets/segmentation/total_segmentator.py
ioangatop: ah yeah just remembered why we are not using cv2; its because the image is in uint16 and the cv2 function does not work with these formats

src/eva/vision/utils/io/nifti.py
nkaenzig: Wouldn't this then raise an error?
`image_data.slicer[:, :, slice_index : slice_index + 1]`

Also the type hint `slice_index: int` would require an update then?

src/eva/vision/data/datasets/segmentation/total_segmentator.py
nkaenzig: to be more specific, why does the concatenation happen along the channel dimension?

src/eva/vision/data/datasets/segmentation/total_segmentator.py
nkaenzig: I see, and why do we need 3 channels?

Nit: Here we use `axis=2` to access the channel dimension, and below in the code we use `axis=-1`

src/eva/vision/utils/io/nifti.py
ioangatop: oh yeah you are right, we have to remove this docs - basically the function reads only slices - updated!

src/eva/vision/data/datasets/segmentation/total_segmentator.py
ioangatop: so each mask - binary mask - is (height, width, 1) which basically represents one label
So here the channel dim is always 1, as the segmentation maps are always binary.

then, to get the multiclass segmentation map, we will need something like (height, width, labels) - I guess that you would probably expecting something like (height, width, 1, labels), but yeah this dim basically is squezed and allows the loss functions to work out-of-the-box CrossEntropyLoss, Dice etc

src/eva/vision/data/datasets/segmentation/total_segmentator.py
ioangatop: yy probably here stuff are more efficient :)

PR #384, opened 2024-04-24
Updated `EmbeddingsWriter` to support multi-embedding file outputs

### LINKED ISSUE #376: Improve grouping feature in `EmbeddingsWriter`
For slide level tasks, we should use `group_key` to group patches from the same slide into the same .pt file(s).
The way it is implemented now still generates one file per patch. We probably should accumulate patches in memory over multiple batches before saving them to file.

nkaenzig:
Closes #376 

- Implemented support for adding multiple embeddings to the same `.pt` if they share the same filename (e.g. slide_id)
- Made `EmbeddingsWriter` agnostic to WSI's by replacing slide_id field extraction by generic `metadata_keys`
- Refactored & updated unit tests

This helps to prevent the explosion of the number of files for slide level tasks, where you can have thousands of patch embeddings per slide. Each slide now results in a single `.pt` file of shape `[n_patches, embedding_dim]`.

### DIFF
--- a/src/eva/core/callbacks/writers/embeddings.py
+++ b/src/eva/core/callbacks/writers/embeddings.py
@@ -3,7 +3,7 @@
 import csv
 import io
 import os
-from typing import Any, Dict, Sequence
+from typing import Any, Dict, List, Sequence
 
 import lightning.pytorch as pl
 import torch
@@ -12,7 +12,7 @@ from loguru import logger
 from torch import multiprocessing, nn
 from typing_extensions import override
 
-from eva.core.callbacks.writers.typings import QUEUE_ITEM
+from eva.core.callbacks.writers.typings import ITEM_DICT_ENTRY, QUEUE_ITEM
 from eva.core.models.modules.typings import DATA_SAMPLE
 from eva.core.utils import multiprocessing as eva_multiprocessing
 
@@ -27,6 +27,7 @@ class EmbeddingsWriter(callbacks.BasePredictionWriter):
         dataloader_idx_map: Dict[int, str] | None = None,
         group_key: str | None = None,
         overwrite: bool = True,
+        save_every_n: int = 1000,
     ) -> None:
         """Initializes a new EmbeddingsWriter instance.
 
@@ -43,6 +44,8 @@ class EmbeddingsWriter(callbacks.BasePredictionWriter):
                 embedding files will be saved in subdirectories named after the group_key.
                 If specified, the key must be present in the metadata of the input batch.
             overwrite: Whether to overwrite the output directory. Defaults to True.
+            save_every_n: Interval for number of iterations to save the embeddings to disk.
+                During this interval, the embeddings are accumulated in memory.
         """
         super().__init__(write_interval="batch")
 
@@ -51,6 +54,7 @@ class EmbeddingsWriter(callbacks.BasePredictionWriter):
         self._dataloader_idx_map = dataloader_idx_map or {}
         self._group_key = group_key
         self._overwrite = overwrite
+        self._save_every_n = save_every_n
 
         self._write_queue: multiprocessing.Queue
         self._write_process: eva_multiprocessing.Process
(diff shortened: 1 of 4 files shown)


src/eva/core/callbacks/writers/embeddings.py
```diff
@@ -122,38 +125,105 @@ def _get_embeddings(self, prediction: torch.Tensor) -> torch.Tensor:
...
+    write_queue: multiprocessing.Queue,
+    output_dir: str,
+    metadata_keys: List[str],
+    save_every_n: int,
+    overwrite: bool = False,
 ) -> None:
-    manifest_file, manifest_writer = _init_manifest(output_dir, overwrite)
+    manifest_file, manifest_writer = _init_manifest(output_dir, metadata_keys, overwrite)
+
+    save_name_to_items: Dict[str, ITEM_DICT_ENTRY] = {}
```
roman807, line 156: not sure if `save_name_to_items` is the best name. It suggests you "save a name to a list of items.."

Can't think of a very good name, but how about `items_dict` or `items_to_save`?

nkaenzig: I like to use variable names in dictionaries that describe the actual mapping, but indeed the `save_` prefix makes it confusing here. What about `name_to_items`?

src/eva/core/callbacks/writers/embeddings.py
```diff
@@ -122,38 +125,105 @@ def _get_embeddings(self, prediction: torch.Tensor) -> torch.Tensor:
...
 
-def _save_prediction(prediction_buffer: io.BytesIO, save_name: str, output_dir: str) -> None:
-    save_path = os.path.join(output_dir, save_name)
-    prediction = torch.load(io.BytesIO(prediction_buffer.getbuffer()), map_location="cpu")
+def _save_items(
+    save_name_to_items: Dict[str, ITEM_DICT_ENTRY],
+    metadata_keys: List[str],
+    output_dir: str,
+    manifest_writer: Any,
+) -> Dict[str, ITEM_DICT_ENTRY]:
```
roman807, line 189: even though this is wip, i would already add a docstring with description. The logic (together with `save_predictions`) is quite tricky and a concise description might make it easier to follow.

src/eva/core/callbacks/writers/embeddings.py
```diff
@@ -122,38 +125,105 @@ def _get_embeddings(self, prediction: torch.Tensor) -> torch.Tensor:
...
+    for save_name, entry in save_name_to_items.items():
+        if len(entry.items) > 0:
+            save_path = os.path.join(output_dir, save_name)
+            is_first_save = entry.save_count == 0
+            if is_first_save:
+                _, target, input_name, _, split, metadata = QUEUE_ITEM(*entry.items[0])
+                metadata = [metadata[key] for key in metadata_keys]  # type: ignore
+                _update_manifest(target, input_name, save_name, split, metadata, manifest_writer)
+            else:
+                pass
```
roman807, line 199: remove the 2 lines above?

src/eva/core/callbacks/writers/embeddings.py
roman807: thats definitely better. Do we follow the <key>_to_<value> naming pattern elsewhere in eva?

nkaenzig: hehe good question, at least I mostly try to use the `to` pattern :D

nkaenzig: done

src/eva/core/callbacks/writers/embeddings.py
nkaenzig: good catch, removed

src/eva/core/callbacks/writers/embeddings.py
nkaenzig: good point, I now added a couple of docstrings to explain what happens.

roman807 (approved):
Great thanks, i think its good to merge

ioangatop (approved):
Looks great!

PR #388, opened 2024-04-25
Replace `cached_property` in `WsiDataset` by LRU cache

### LINKED ISSUE #387: Solve memory overflow issue with ` MultiWsiDataset`
Memory usage seems to continuously grow while iterating through the dataset:

image

Most likely due to the cached properties which accumulate in memory.

nkaenzig:
Closes #387 

Tested using 10k WSIs with 100 patches each using a dataloader with 10 workers. With this fix the memory doesn't build up anymore:
image


roman807 (approved):
LGTM, thanks!

PR #390, opened 2024-04-25
Update `TotalSegmentator` output binary masks to a semantic mask

### LINKED ISSUE #389: Updates of `TotalSegmentator` dataset target masks  to segmentation label masks
In the masks outputs, we should include the background mask and instead on one-hot-encoded binary masks (NHW), to return a segmentation label masks (HW) with each integer to show the label

ioangatop:
Closes #389

### DIFF
--- a/src/eva/vision/data/datasets/segmentation/total_segmentator.py
+++ b/src/eva/vision/data/datasets/segmentation/total_segmentator.py
@@ -131,13 +131,18 @@ class TotalSegmentator2D(base.ImageSegmentation):
         return tv_tensors.Image(image_rgb_array.transpose(2, 0, 1))
 
     @override
-    def load_masks(self, index: int) -> tv_tensors.Mask:
+    def load_mask(self, index: int) -> tv_tensors.Mask:
         masks_dir = self._get_masks_dir(index)
         slice_index = self._get_sample_slice_index(index)
         mask_paths = (os.path.join(masks_dir, label + ".nii.gz") for label in self.classes)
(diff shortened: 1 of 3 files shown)


src/eva/vision/data/datasets/segmentation/total_segmentator.py
```diff
@@ -131,13 +131,18 @@ def load_image(self, index: int) -> tv_tensors.Image:
...
-        list_of_mask_arrays = [io.read_nifti_slice(path, slice_index) for path in mask_paths]
-        masks = np.concatenate(list_of_mask_arrays, axis=2)
-        return tv_tensors.Mask(masks.transpose(2, 0, 1))
+        one_hot_encoded = np.concatenate(
+            [io.read_nifti_slice(path, slice_index) for path in mask_paths],
+            axis=2,
+        )
+        background_mask = one_hot_encoded.sum(axis=2, keepdims=True) == 0
+        one_hot_encoded_with_bg = np.concatenate([background_mask, one_hot_encoded], axis=2)
+        segmentation_label = np.argmax(one_hot_encoded_with_bg, axis=2)
```
nkaenzig, line 144: Why doing the concat with the background mask followed by argmax?
Might be worth to add a comment / docstring here to make this easier to understand.

ioangatop: Good question, we asked the same question! No, they are not overlapping and we use it as a 2d semantic segmentation task

nkaenzig (approved):
LGTM

Issue #393, opened 2024-04-25
Implement simple baseline sampler for foreground patches


roman807:
See reference implementation: https://www.kaggle.com/code/akensert/panda-optimized-tiling-tf-data-dataset/notebook

PR #394, opened 2024-04-25
Simple baseline sampler for foreground patches

roman807:
closes https://github.com/kaiko-ai/eva/issues/393

### DIFF
--- a/src/eva/core/data/datasets/embeddings/classification/multi_embeddings.py
+++ b/src/eva/core/data/datasets/embeddings/classification/multi_embeddings.py
@@ -83,6 +83,12 @@ class MultiEmbeddingsClassificationDataset(base.EmbeddingsDataset):
 
         if not embeddings.ndim == 2:
             raise ValueError(f"Expected 2D tensor, got {embeddings.ndim} for {multi_id}.")
+        
+        # pad to [100, 384] if necessary
+        n_embeddings = 150 # 200
+        if embeddings.shape[0] < n_embeddings:
+            n_pad = n_embeddings - embeddings.shape[0]
+            embeddings = torch.nn.functional.pad(embeddings, (0,0,0,n_pad), mode='constant', value=0)
 
         return embeddings
 
--- a/src/eva/core/data/datasets/embeddings/classification/multi_embeddings.py
+++ b/src/eva/core/data/datasets/embeddings/classification/multi_embeddings.py
@@ -21,6 +21,7 @@ class MultiEmbeddingsClassificationDataset(base.EmbeddingsDataset):
         root: str,
         manifest_file: str,
         split: Literal["train", "val", "test"],
+        n_embeddings: int,
         column_mapping: Dict[str, str] = base.default_column_mapping,
         embeddings_transforms: Callable | None = None,
         target_transforms: Callable | None = None,
@@ -42,6 +43,8 @@ class MultiEmbeddingsClassificationDataset(base.EmbeddingsDataset):
                 the `root` argument.
             split: The dataset split to use. The `split` column of the manifest
                 file will be splitted based on this value.
+            n_embeddings: Expected number of embeddings per sample. If less, the embeddings
+                will be padded with zeros.
             column_mapping: Defines the map between the variables and the manifest
                 columns. It will overwrite the `default_column_mapping` with
                 the provided values, so that `column_mapping` can contain only the
@@ -59,6 +62,7 @@ class MultiEmbeddingsClassificationDataset(base.EmbeddingsDataset):
         )
 
         self._multi_ids: List[int]
+        self._n_embeddings = n_embeddings
 
     @override
     def setup(self):
@@ -83,12 +87,12 @@ class MultiEmbeddingsClassificationDataset(base.EmbeddingsDataset):
 
         if not embeddings.ndim == 2:
             raise ValueError(f"Expected 2D tensor, got {embeddings.ndim} for {multi_id}.")
-        
-        # pad to [100, 384] if necessary
-        n_embeddings = 150 # 200
-        if embeddings.shape[0] < n_embeddings:
-            n_pad = n_embeddings - embeddings.shape[0]
-            embeddings = torch.nn.functional.pad(embeddings, (0,0,0,n_pad), mode='constant', value=0)
+
+        if embeddings.shape[0] < self._n_embeddings:
+            n_pad = self._n_embeddings - embeddings.shape[0]
+            embeddings = torch.nn.functional.pad(
+                embeddings, (0, 0, 0, n_pad), mode="constant", value=0
+            )
 
         return embeddings
 
--- a/src/eva/core/data/datasets/embeddings/classification/multi_embeddings.py
+++ b/src/eva/core/data/datasets/embeddings/classification/multi_embeddings.py
@@ -21,7 +21,6 @@ class MultiEmbeddingsClassificationDataset(base.EmbeddingsDataset):
         root: str,
         manifest_file: str,
         split: Literal["train", "val", "test"],
-        n_embeddings: int,
         column_mapping: Dict[str, str] = base.default_column_mapping,
         embeddings_transforms: Callable | None = None,
         target_transforms: Callable | None = None,
@@ -62,7 +61,6 @@ class MultiEmbeddingsClassificationDataset(base.EmbeddingsDataset):
         )
 
         self._multi_ids: List[int]
-        self._n_embeddings = n_embeddings
 
     @override
     def setup(self):
@@ -87,12 +85,6 @@ class MultiEmbeddingsClassificationDataset(base.EmbeddingsDataset):
 
         if not embeddings.ndim == 2:
             raise ValueError(f"Expected 2D tensor, got {embeddings.ndim} for {multi_id}.")
-
-        if embeddings.shape[0] < self._n_embeddings:
-            n_pad = self._n_embeddings - embeddings.shape[0]
-            embeddings = torch.nn.functional.pad(
-                embeddings, (0, 0, 0, n_pad), mode="constant", value=0
-            )
 
         return embeddings
 
--- a/src/eva/vision/data/datasets/wsi.py
+++ b/src/eva/vision/data/datasets/wsi.py
@@ -1,7 +1,6 @@
 """Dataset classes for whole-slide images."""
 
 import os
-from functools import cached_property
 from typing import Callable, Dict
 
 import pandas as pd
@@ -11,7 +10,6 @@ from typing_extensions import override
 
 from eva.vision.data import wsi
 from eva.vision.data.datasets import vision
-from eva.vision.data.wsi.backends import wsi_backend
 from eva.vision.data.wsi.patching import samplers
 
 
@@ -55,20 +53,19 @@ class WsiDataset(vision.VisionDataset):
     def filename(self, index: int) -> str:
         return f"{self._file_path}_{index}"
 
-    @cached_property
+    @property
     def _wsi(self) -> wsi.Wsi:
-        # wsi_obj = wsi.get_wsi_class(self._backend)(self._file_path)
-        return wsi_backend(self._backend)(self._file_path)
+        return wsi.get_cached_wsi(self._file_path, self._backend)
 
-    @cached_property
+    @property
     def _coords(self) -> wsi.PatchCoordinates:
-        return wsi.PatchCoordinates.from_file(
-            wsi_path=self._file_path,
+        return wsi.get_cached_coords(
+            file_path=self._file_path,
             width=self._width,
             height=self._height,
             target_mpp=self._target_mpp,
-            backend=self._backend,
             sampler=self._sampler,
+            backend=self._backend,
         )
 
     @override
--- a/src/eva/vision/data/datasets/wsi.py
+++ b/src/eva/vision/data/datasets/wsi.py
@@ -72,7 +72,7 @@ class WsiDataset(vision.VisionDataset):
     def __getitem__(self, index: int) -> torch.Tensor:
         x, y = self._coords.x_y[index]
         width, height, level_idx = self._coords.width, self._coords.height, self._coords.level_idx
-        patch = self._wsi.read_region((x, y), (width, height), level_idx)
+        patch = self._wsi.read_region((x, y), level_idx, (width, height))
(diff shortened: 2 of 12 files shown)


src/eva/core/data/datasets/embeddings/classification/multi_embeddings.py
```diff
@@ -84,6 +88,12 @@ def _load_embeddings(self, index: int) -> torch.Tensor:
         if not embeddings.ndim == 2:
             raise ValueError(f"Expected 2D tensor, got {embeddings.ndim} for {multi_id}.")
 
+        if embeddings.shape[0] < self._n_embeddings:
```
nkaenzig, line 91: The padding logic is not necessary here, we can do that via the `embedding_transforms`.
See example here:
https://github.com/kaiko-ai/eva/blob/767fbbd399961560e0796e4c31187225584c791e/tests/eva/core/data/datasets/embeddings/classification/test_multi_embeddings.py#L88

src/eva/vision/data/wsi/patching/coordinates.py
```diff
@@ -55,8 +57,32 @@ def from_file(
         mpp_ratio = target_mpp / level_mpp
         scaled_width, scaled_height = int(mpp_ratio * width), int(mpp_ratio * height)
 
-        for x, y in sampler.sample(scaled_width, scaled_height, wsi.level_dimensions[level_idx]):
```
nkaenzig, line 58: I would leave the code here unchanged, and let the sampler be in charge to only return the x,y 's that are relevant.

To achieve that, here a high level description of how I would implement this feature:

1. Add a `_mask` `@property` to `WsiDataset` which returns a numpy array that represents the mask (at the maximum wsi level (the one with lowest resolution))
2. Add optional `mask` and `min_foreground` arguments to `PatchCoordinates`, `RandomSampler` and `GridSampler` (I wouldn't write custom samplers for this, because the additional logic for doing the foreground filtering can be very minimal here, and using masks will be the default case anyways)
3. Then in the samplers, yield only the x,y coordinates of patches that exceed the `min_foreground` threshold, this would look similar to the following code for the RandomSampler as an example

```
for _ in range(self.n_samples):
    x_max, y_max = layer_shape[0], layer_shape[1]
    x, y = random.randint(0, x_max - width), random.randint(0, y_max - height)  # nosec
    patch_mask = mask[x:x + width, y:y + height]
    if mask is not None:
        if min_foreground > patch_mask.sum() / patch_mask.size:
            yield x, y
    else:
        yield x, y 
```

In `mask[x:x + width, y:y + height]` you also might need to scale the coords & width from the max level to the level used here.

What do you think? Am I missing something?

nkaenzig (commented):
Looks good, I left some comments.

src/eva/vision/data/wsi/patching/coordinates.py
```diff
@@ -55,8 +58,28 @@ def from_file(
...
-            x_y.append((x, y))
+        if not isinstance(sampler, samplers.ForegroundGridSampler):
```
nkaenzig, line 61: I'd replace this by something like this:

```
if isinstance(sampler, samplers.ForegroundSampler):
    mask = get_mask(...)  # do the scaling in this function as well

    for x, y in sampler.sample(
        scaled_width, scaled_height, wsi.level_dimensions[level_idx], mask=mask
     ):
        x_y.append((x, y))
```

We can add `*args` to the init args of the samplers, so passing `mask=mask` argument won't do anything, or we can build a dict `sample_args` here and call `sample(*sample_args)` , and add the `mask` key only for foreground samplers. Maybe the latter is cleaner?

Then in `samplers.py` we can add an abstract base class for foreground samplers (as eventually we might have multiple of those)

```
class ForegroundSampler(abc.ABC, Sampler):
    """Base class for samplers with foreground filtering capabilities."""

    @abc.abstractmethod
    def sample(
        self,
        width: int,
        height: int,
        layer_shape: tuple[int, int],
        mask: np.ndarray,
    ) -> Generator[Tuple[int, int], None, None]:
        """Iterator that samples patches."""

    @abc.abstractmethod
    def is_foreground(...):
        pass
```

src/eva/vision/utils/io/image.py
```diff
@@ -52,3 +52,30 @@ def read_image_as_array(path: str, flags: int = cv2.IMREAD_UNCHANGED) -> npt.NDA
         image = image[:, :, np.newaxis]
 
     return np.asarray(image).astype(np.uint8)
+
+
+def get_mask(
```
nkaenzig, line 57: I'm not sure if this belongs here, as it's not really IO?
Maybe in `src/eva/vision/data/wsi/masking/kaggle.py` (replace kaggle by a better name)

src/eva/vision/data/wsi/patching/samplers.py
```diff
@@ -93,13 +95,83 @@ def sample(
...
+
+    def is_foreground(
+        self,
+        mask: np.ndarray,
+        x: int,
+        y: int,
+        width: int,
+        height: int,
+        mask_scale_factor: float,
+        min_patch_info=0.35,
```
nkaenzig, line 156: maybe rename to `min_foreground_ratio` or similar?

src/eva/vision/data/wsi/patching/coordinates.py
roman807: refactored accordingly. However, note that now the ForegroundGridSampler doesn't inherit from GridSampler (due to issue with signature when inheriting from both, ForegroundSampler and GridSampler). Moved the shared grid-sampling logic to the function `_get_grid_coords_and_indices` instead

src/eva/vision/data/wsi/patching/coordinates.py
```diff
@@ -49,13 +50,21 @@ def from_file(
...
+        sample_args = {}
+        sample_args["width"] = scaled_width
+        sample_args["height"] = scaled_height
+        sample_args["layer_shape"] = wsi.level_dimensions[level_idx]
+
+        if isinstance(sampler, samplers.ForegroundSampler):
+            sample_args["mask"] = get_mask(wsi, wsi_path, level_idx)
+
+        x_y = []
+        for x, y in sampler.sample(**sample_args):
```
nkaenzig, line 67:
```suggestion
        sample_args = {
            "width": scaled_width,
            "height": scaled_height,
            "layer_shape": wsi.level_dimensions[level_idx]
        }

        if isinstance(sampler, samplers.ForegroundSampler):
            sample_args["mask"] = get_mask(wsi, wsi_path, level_idx)

        x_y = [(x, y) for x, y in sampler.sample(**sample_args)]
```

src/eva/vision/data/wsi/patching/coordinates.py
```diff
@@ -49,13 +50,21 @@ def from_file(
...
             x_y.append((x, y))
```
nkaenzig, line 68:
```suggestion
        sample_args = {
            "width": scaled_width,
            "height": scaled_height,
            "layer_shape": wsi.level_dimensions[level_idx]
        }

        if isinstance(sampler, samplers.ForegroundSampler):
            sample_args["mask"] = get_mask(wsi, wsi_path, level_idx)

        x_y = [(x, y) for x, y in sampler.sample(**sample_args)]
```

src/eva/vision/utils/io/image.py
```diff
@@ -52,3 +53,40 @@ def read_image_as_array(path: str, flags: int = cv2.IMREAD_UNCHANGED) -> npt.NDA
...
+    wsi: Any,
```
nkaenzig, line 59: Can use `eva.vision.data.wsi.backends.base.Wsi` as type hint here

src/eva/vision/utils/io/image.py
```diff
@@ -52,3 +53,40 @@ def read_image_as_array(path: str, flags: int = cv2.IMREAD_UNCHANGED) -> npt.NDA
...
+    """Extracts a binary mask from an image.
+    
+    Args:
+        image: The input image.
+        kernel_size: The size of the kernel for morphological operations.
+        gray_threshold: The threshold for the gray scale image.
+        fill_holes: Whether to fill holes in the mask.
+    """
+    image = np.array(
+        wsi.open_file(wsi_path).read_region(
```
nkaenzig, line 75: `open_file` here shouldn't be necessary here, `wsi` is already an initialized object (see `backends.wsi_backend(backend)(wsi_path)` in coordinates.py), so here, you can call `wsi.read_region(..)` instead

src/eva/vision/utils/io/image.py
```diff
@@ -52,3 +53,40 @@ def read_image_as_array(path: str, flags: int = cv2.IMREAD_UNCHANGED) -> npt.NDA
         image = image[:, :, np.newaxis]
 
     return np.asarray(image).astype(np.uint8)
+
+
+def get_mask(
+    wsi: Any,
+    wsi_path: str,
```
nkaenzig, line 60: `wsi_path` can be removed here (see comment below)

src/eva/vision/utils/io/image.py
```diff
@@ -52,3 +53,40 @@ def read_image_as_array(path: str, flags: int = cv2.IMREAD_UNCHANGED) -> npt.NDA
...
+    level_idx: int,
+    kernel_size: tuple[int, int] = (7, 7),
+    gray_threshold: int = 220,
+    fill_holes: bool = False,
+) -> tuple[np.ndarray, float]:
+    """Extracts a binary mask from an image.
+    
+    Args:
+        image: The input image.
```
nkaenzig, line 69: Update docstring

src/eva/vision/data/wsi/patching/coordinates.py
nkaenzig: I see, I think it looks good now, thanks for addressing this. 🚀

src/eva/vision/utils/io/image.py
roman807: Thanks, adjusting this. i also noticed in our implementation of `src/eva/vision/data/wsi/backends/openslide.py` we mixed up the order of the arguments (`location, size, level` vs `location, level, size` in openslide), changing it to `location, level, size` everywhere

nkaenzig: Good catch!

src/eva/vision/utils/io/image.py
roman807: how about a `mask.py` in `src/eva/vision/utils`? I'm thinking eventually we'll only have one get_mask function (if we pick what works best, there should be no need for alternatives)

src/eva/vision/data/wsi/patching/samplers.py
```diff
@@ -93,13 +95,83 @@ def sample(
...
+                yield x, y
+
+    def is_foreground(
+        self,
+        mask: np.ndarray,
+        x: int,
+        y: int,
+        width: int,
+        height: int,
+        mask_scale_factor: float,
```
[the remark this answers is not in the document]
roman807: not sure if this is worth it: it would mean (1) resize the mask (linear interpolation) + (2) computing the foreground ratio for a much larger patch. Basically extra computation just for nicer code.(?)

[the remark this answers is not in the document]
nkaenzig: Good points, let's leave it as is

src/eva/vision/utils/io/image.py
nkaenzig: `src/eva/vision/utils/mask.py` I think would be fine for now (I'd move it out of coordinates). We can later move / refactor this part when we want to support multiple masking methods.

nkaenzig (approved):
Looks great! Maybe @ioangatop could have a last look before we merge?

roman807:
> Looks great! Maybe @ioangatop could have a last look before we merge?

sorry saw this only after merging. But feel free to comment in this pr and i'll open a new issue

ioangatop:
>> Looks great! Maybe @ioangatop could have a last look before we merge?
> sorry saw this only after merging. But feel free to comment in this pr and i'll open a new issue

no problem, yes lets do that

src/eva/vision/data/datasets/wsi.py
```diff
@@ -72,7 +72,7 @@ def _coords(self) -> wsi.PatchCoordinates:
     def __getitem__(self, index: int) -> torch.Tensor:
         x, y = self._coords.x_y[index]
         width, height, level_idx = self._coords.width, self._coords.height, self._coords.level_idx
-        patch = self._wsi.read_region((x, y), (width, height), level_idx)
+        patch = self._wsi.read_region((x, y), level_idx, (width, height))
```
ioangatop, line 75: here its good to add the kwargs names, like `self._wsi.read_region((x, y), level=level_idx, size=(width, height))`

src/eva/vision/data/wsi/patching/samplers.py
```diff
@@ -87,19 +115,112 @@ def sample(
...
+        self.overlap = overlap
+        self.seed = seed
+
+    def sample(
+        self,
+        width: int,
+        height: int,
+        layer_shape: tuple[int, int],
+        mask: tuple[np.ndarray, float],
+    ):
```
ioangatop, line 148: lets add return type here

src/eva/vision/data/wsi/patching/samplers.py
```diff
@@ -87,19 +115,112 @@ def sample(
...
+
+
+def _get_grid_coords_and_indices(
+    layer_shape: tuple[int, int],
+    width: int,
+    height: int,
+    overlap: tuple[int, int],
+    shuffle: bool = True,
+    seed: int = 42,
+):
```
ioangatop, line 204: return type here

src/eva/vision/utils/mask.py
```diff
@@ -0,0 +1,39 @@
...
+from eva.vision.data.wsi.backends.base import Wsi
+
+
+def get_mask(
+    wsi: Wsi,
+    level_idx: int,
+    kernel_size: tuple[int, int] = (7, 7),
+    gray_threshold: int = 220,
+    fill_holes: bool = False,
+) -> tuple[np.ndarray, float]:
```
ioangatop, line 13: lets use `from typing import Tuple; Tuple[..]` to be consistent; this also helps to support older python versions, but at some point maybe we want to change everything to lowcase

ioangatop (commented):
Looks awesome, some minor comments that the CI should caught (if it run 😅 )

PR #395, opened 2024-04-25
Add timm encoder and model wrapper networks

### LINKED ISSUE #381: Add `timm` encoder and model wrapper networks
We would like to add a `timm` model wrapper and encoder, which will be able to extract the intermediate layers of backbones, in order to pass them to a segmentation decoder

ioangatop:
Closes #381

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,14 +59,14 @@ vision = [
     "h5py>=3.10.0",
     "nibabel>=5.2.0",
     "opencv-python-headless>=4.9.0.80",
-    "timm>=0.9.12",
+    "timm @ git+https://github.com/huggingface/pytorch-image-models.git@main",
     "torchvision>=0.17.0",
 ]
 all = [
     "h5py>=3.10.0",
     "nibabel>=5.2.0",
     "opencv-python-headless>=4.9.0.80",
-    "timm>=0.9.12",
+    "timm @ git+https://github.com/huggingface/pytorch-image-models.git@main",
     "torchvision>=0.17.0",
 ]
 
--- a/src/eva/vision/models/networks/encoders/encoder.py
+++ b/src/eva/vision/models/networks/encoders/encoder.py
@@ -1,0 +1,23 @@
+"""Encoder base class."""
+
+import abc
+from typing import List
+
+import torch
+from torch import nn
+
+
+class Encoder(nn.Module, abc.ABC):
+    """Encoder base class."""
+
+    @abc.abstractmethod
+    def forward(self, tensor: torch.Tensor) -> List[torch.Tensor]:
+        """Returns the multi-level feature maps of the model.
+
+        Args:
+            tensor: The image tensor (batch_size, num_channels, height, width).
+
+        Returns:
+            The list of multi-level image features of shape (batch_size,
+                hidden_size, num_patches_height, num_patches_width).
+        """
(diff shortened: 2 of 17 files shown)


pyproject.toml
```diff
@@ -59,14 +59,14 @@ vision = [
     "h5py>=3.10.0",
     "nibabel>=5.2.0",
     "opencv-python-headless>=4.9.0.80",
-    "timm>=0.9.12",
+    "timm @ git+https://github.com/huggingface/pytorch-image-models.git@main",
```
nkaenzig, line 62: Should we open an issue to set this back to a specific version tag once there is a new release out?

src/eva/vision/models/networks/from_timm.py
```diff
@@ -0,0 +1,49 @@
...
+        self._model_name = model_name
+        self._pretrained = pretrained
+        self._checkpoint_path = checkpoint_path
+        self._model_arguments = model_arguments or {}
+
+        self._feature_extractor = self._load_model()
+
+    def _load_model(self) -> nn.Module:
+        """Builds, loads and returns the model."""
+        return timm.create_model(
```
nkaenzig, line 40: Couldn't we use use the existing `ModelFromFunction` in conjunction with `timm.create_model`?

src/eva/vision/models/networks/from_timm.py
```diff
@@ -0,0 +1,49 @@
...
+
+from typing import Any, Dict, List
+
+import timm
+import torch
+from torch import nn
+from typing_extensions import override
+
+
+class TimmModel(nn.Module):
```
nkaenzig, line 11: Why not using the existing wrapper API here?

src/eva/vision/models/networks/encoders/encoder.py
```diff
@@ -0,0 +1,23 @@
+"""Encoder base class."""
+
+import abc
+from typing import List
+
+import torch
+from torch import nn
+
+
+class Encoder(nn.Module, abc.ABC):
```
nkaenzig, line 10: What is the main motivation for introducing the new encoder / decoder abstractions at this point?
Previously we had backbone / head - how do these play together?

nkaenzig (commented):
Looks good! I left some comments.

pyproject.toml
ioangatop: sure yeah - maybe it will be caught from the bot as well 🤔

src/eva/vision/models/networks/encoders/encoder.py
ioangatop: in out case backbone outputs a single tensor which encoder a list of them (and the head / decoder process the inputs accordingly)

maybe we could have a conversation about these conventions

src/eva/vision/models/networks/from_timm.py
ioangatop: I though maybe it would be easier for the users, which the majority of them will use the `TimmModels` to have a dedicated wrapper, since the `ModelFromFunction` will make them undestund that the model from timm initialized with a function, in yaml we can not init a function that's why we have to use the wrapper `ModelFromFunction`

what do you think?

nkaenzig: Good point. I think it's fine to have a dedicated wrapper for Timm, but we should use the wrapper api from eva.core I think?

ioangatop: oh yes no problem!

ioangatop: done

PR #403, opened 2024-04-29
Add `timm` encoder networks

### LINKED ISSUE #381: Add `timm` encoder and model wrapper networks

ioangatop:
Closes #381


src/eva/vision/models/networks/encoders/from_timm.py
```diff
@@ -0,0 +1,61 @@
...
+            model_name=self._model_name,
+            pretrained=self._pretrained,
+            checkpoint_path=self._checkpoint_path,
+            out_indices=self._out_indices,
+            features_only=True,
+            **self._model_arguments,
+        )
+
+    @override
+    def forward(self, tensor: torch.Tensor) -> List[torch.Tensor]:
```
ioangatop, line 60: @nkaenzig I dont know if we can use `wrappers.ModelFromFunction` here because the return type is a `List[Tensor]` and not a `Tensor`

src/eva/vision/models/networks/from_timm.py
nkaenzig, line 1: Should we put this into a `wrappers` subdir (following the folder structure used in `core`)?

src/eva/vision/models/networks/encoders/from_timm.py
nkaenzig, line 1: Should we change the filename here? `from_timm` to `timm` only?

nkaenzig (approved):
LGTM

src/eva/vision/models/networks/encoders/from_timm.py
ioangatop: we cant use timm because then it will be a conflict with `import timm` :/ - I'm open to other suggestions though!

src/eva/vision/models/networks/from_timm.py
ioangatop: i'm fine with it but also I think we can do it once we have more than 🤔

src/eva/vision/models/networks/encoders/from_timm.py
nkaenzig: Right, `timm_encoder` maybe, but that violates our naming convention as it's redundant with the `encoders/` parent dir. Can't think of something better either, maybe you can, but elsewise let's leave as is.

ioangatop: `timm_model`?

PR #404, opened 2024-04-29
Add simple decoder networks for segmentation tasks

### LINKED ISSUE #167: Add linear and conv2d decoder for segmentation tasks
Lets add a linear and conv2d decoder heads for 2D segmentation tasks.

This will be along with a segmentation config for `TotalSegmentor2D` task.

ioangatop:
Closes #167


src/eva/vision/models/networks/decoders/decoder.py
nkaenzig, line 1: Following the conversation from here: https://github.com/kaiko-ai/eva/pull/396#discussion_r1582719840

What if we remove segmentation here from the docstrings, keep `decoder.py` here, and move `conv.py` and `linear.py` to a segmentation subfolder? This would make it clearer that these are intended for segmentation only.

src/eva/vision/models/networks/decoders/linear.py
```diff
@@ -0,0 +1,108 @@
...
+
+        Args:
+            patch_embeddings: The patch embeddings tensor of shape
+                (batch_size, hidden_size, n_patches_height, n_patches_width).
+
+        Returns:
+            The logits as a tensor (batch_size, n_classes, n_patches_height,
+            n_patches_width).
+        """
+        batch_size, _, height, width = patch_embeddings.shape
```
nkaenzig, line 67: Consider using `hidden_size` explicitely instead of `_` placeholder, or is there a benefit of using `_` here?

nkaenzig (approved):
Looks great. Left two last minor comments.

src/eva/vision/models/networks/decoders/linear.py
ioangatop: nice done

src/eva/vision/models/networks/decoders/decoder.py
ioangatop: that's a very good idea! done

PR #409, opened 2024-04-30
Move model freezing functionality to `configure_model`

### LINKED ISSUE #408: Freeze model in `configuration_model` method in `HeadModule`
We would like to freeze the model in the `configuration_model` step, so that the information in the logs is accurate

Screenshot 2024-04-30 at 1 25 20 PM

while instead we would like to have something like the following:

Screenshot 2024-04-30 at 1 31 21 PM

ioangatop:
Closes #408


nkaenzig (approved):
LGTM

PR #412, opened 2024-05-03
Add lower bound for wsi resolution level during mask generation

### LINKED ISSUE #411: Fix issue where `ForegroundGridSampler` doesn't return any patches
The current implementation which calculates the mask takes the maximum (lowest resolution level) from the wsi and calculates the mask for that. For some WSIs, the dimensions of the maximum level are too small, thus the scaled x, y coordinates of the patches collapse to a single pixel.

nkaenzig:
Closes #411


src/eva/vision/data/wsi/patching/mask.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    valid_level_index = len(wsi.level_dimensions) - 1
+
+    if min_pixels is None:
+        return valid_level_index
+    else:
+        for index, (width, height) in reversed(list(enumerate(wsi.level_dimensions))):
+            if width * height >= min_pixels:
+                valid_level_index = index
+                break
```
roman807, line 76: should we throw an error message here if min_pixels has been set too high and we don't get a valid_level_index?

nkaenzig: Good point. I now added this.

roman807 (approved):

PR #420, opened 2024-05-07
Move sampler logic to `samplers` module and add unit tests

### LINKED ISSUE #419: Write unit tests for WSI patch sampler classes

nkaenzig:
Closes #419


src/eva/vision/data/wsi/patching/samplers/grid.py
```diff
@@ -0,0 +1,47 @@
+"""Grid sampler."""
+
+from typing import Generator, Tuple
+
+from eva.vision.data.wsi.patching.samplers import _utils, base
+
+
+class GridSampler(base.Sampler):
```
roman807, line 8: do you think there will be need for this one?

nkaenzig: Maybe for datasets such as BACH this might come in useful (not for the slide level task, but the patch level dataset as the patches there are too big and downscaling them to 224 as we currently do leads to loss of information and mpp's the FMs are usually not trained on much).

roman807 (approved):

PR #429, opened 2024-05-08
Add `WsiClassificationDataset`

### LINKED ISSUE #428: Add a task-independent dataset class for WSI classification tasks
Next to the dataset specific dataset classes, it would be handy to have a task-independent WSI dataset class that supports manifest files as input.

nkaenzig:
Closes #428


roman807:
Thanks @nkaenzig! Could you also update the `configs/vision/dino_vit/offline/panda.yaml`?

src/eva/vision/data/datasets/classification/wsi.py
```diff
@@ -1,40 +1,109 @@
...
+            self,
+            root=root,
+            file_paths=self._file_paths,
+            width=width,
+            height=height,
+            sampler=sampler,
+            target_mpp=target_mpp,
+            backend=backend,
+            image_transforms=image_transforms,
+        )
```
ioangatop, line 77: maybe we should call super at the start and then register (self._*) the attributes that are not in the parent class; like I guess `self._root` should be available in the parent class right?

ioangatop (approved):
looks great! just a minor comment

src/eva/vision/data/datasets/classification/wsi.py
nkaenzig: Great catch! I updated this now (I had to register some attributes before the super call though, because I have to load the manifest / list of file paths first)

PR #430, opened 2024-05-08
Add `PANDA` dataset class

### LINKED ISSUE #425: Create a dataset class for `PANDA`

nkaenzig:
Closes #425


src/eva/vision/data/datasets/classification/panda.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+class PANDA(wsi.MultiWsiDataset, base.ImageClassification):
+    """Dataset class for PANDA images and corresponding targets."""
+
+    _train_split_ratio: float = 0.8
+    """Train split ratio."""
+
+    _val_split_ratio: float = 0.2
+    """Validation split ratio."""
```
roman807, line 26: these ratios don't correspond to what we have in the readme (0.7/0.15/0.15). Also a test split is missing. (Here we have sufficient data to create a test split so we should probably create and use it)

nkaenzig: Good point, I wanted to discuss this with you. For panda there is also the leaderboard on kaggle, which will serve as the real test set. But still, for offline evaluation I agree that it makes sense to have train/val/test as well.

roman807: yeah, evaluating and reporting on a val set (which was used for early stopping) is not ideal and should only be done if there is no other way (i.e. splits are given by provider / no option to split three-ways without data leakage etc.). Here the splits are created by us and there is sufficient data, so i think there is no reason to not have a test set.

nkaenzig: Updated this now to generate stratified train/val/test splits.

configs/vision/dino_vit/offline/panda.yaml
```diff
@@ -84,8 +84,7 @@ data:
           embeddings_transforms:
             class_path: eva.core.data.transforms.Pad2DTensor
             init_args:
-              pad_size: 100
-              pad_value: 0
+              pad_size: &N_PATCHES 100
```
roman807, line 87: set this to 1000 instead?

docs/datasets/panda.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Key stats
+
+|                           |                                                          |
+|---------------------------|----------------------------------------------------------|
+| **Modality**              | Vision (WSI)                                             |
+| **Task**                  | Multiclass classification (6 classes)                    |
+| **Cancer type**           | Prostate                                                 |
+| **Data size**             | 347 GB                                                   |
+| **Image dimension**       | 1536 x 2048 x 3                                          |
```
roman807, line 20: change this to `average ~20-25k x ~20-25k x 3`

docs/datasets/panda.md
```diff
@@ -0,0 +1,68 @@
+# PANDA (Prostate cANcer graDe Assessment)
+
+The PANDA datasets consists of 10616 whole-slide images of digitized H&E-stained prostate tissue biopsies originating from two medical centers. After the biopsy, the slides were classified into Gleason patterns (3, 4 or 5) based on the architectural growth patterns of the tumor, which are then converted into an ISUP grade on a 1-5 scale.
+
```
roman807, line 4: `...0-5 scale`

docs/datasets/panda.md
```diff
@@ -0,0 +1,68 @@
...
+│   ├── 0005f7aaab2800f6170c399693a96917_mask.tiff
+│   └── ...
+├── train.csv (contains Gleason & ISUP labels)
+├── test.csv
+├── sample_submission.csv
+```
+
+## Download and preprocessing
+
+The `PANDA` dataset class doesn't downloading the data during runtime and must be downloaded manually from [kaggle](https://www.kaggle.com/competitions/prostate-cancer-grade-assessment/data).
```
roman807, line 45: `downloading` -> `download`

src/eva/core/data/splitting/stratified.py
```diff
@@ -0,0 +1,52 @@
...
+    n_classes = unique_classes.shape[0]
+
+    train_indices, val_indices, test_indices = [], [], []
+
+    for c in range(n_classes):
+        class_indices = np.where(y_indices == c)[0]
+        np.random.shuffle(class_indices)
+
+        n_train = int(np.floor(train_ratio * len(class_indices))) or 1
+        n_val = int(np.floor(val_ratio * len(class_indices))) or 1
```
roman807, line 45: `n_train`+`n_val` might not add up to `len(class_indices)`. what about something like:
```
if test_ratio is None or test_ratio <= 0.0:
   n_val = len(class_indices) - len(class_indices)
else:
  n_val = int(np.floor(val_ratio * len(class_indices))) or 1
train_indices.extend(class_indices[:n_train])
val_indices.extend(class_indices[n_train : n_train + n_val])
test_indices.extend(class_indices[n_train + n_val :])
```

src/eva/core/data/splitting/stratified.py
```diff
@@ -0,0 +1,52 @@
...
+
+import numpy as np
+
+
+def stratified_split(
+    samples: Sequence[Any],
+    targets: Sequence[Any],
+    train_ratio: float,
+    val_ratio: float,
+    test_ratio: float | None = None,
```
roman807, line 13: maybe just default test_ratio to 0, then we don't need `test_ratio is not None and test_ratio > 0.0` later

docs/datasets/panda.md
nkaenzig: done

docs/datasets/panda.md
nkaenzig: done

docs/datasets/panda.md
nkaenzig: done

src/eva/core/data/splitting/stratified.py
nkaenzig: good point. I updated this now accordingly

src/eva/core/data/splitting/stratified.py
nkaenzig: good catch. fixed now

roman807 (approved):

roman807:
@nkaenzig Please also add the panda dataset to the table in https://github.com/kaiko-ai/eva/blob/425-create-panda-dataset-class/docs/datasets/index.md

src/eva/vision/data/datasets/classification/panda.py
```diff
@@ -0,0 +1,188 @@
...
+        is_noisy_filter = self.annotations["noise_ratio_10"] == 0
+        non_noisy_image_ids = set(self.annotations.loc[~is_noisy_filter].index)
+        filtered_file_paths = [
+            file_path
+            for file_path in file_paths
+            if self._get_id_from_path(file_path) in non_noisy_image_ids
+        ]
+        return filtered_file_paths
+
+    def _path_to_target(self, file_path: str) -> int:
```
roman807, line 184: could we rename this to `_get_target_from_path` or similar? the current name suggests we look for a path to a target

src/eva/vision/data/datasets/classification/panda.py
```diff
@@ -0,0 +1,188 @@
+"""Dataset classes for whole-slide image classification."""
```
roman807, line 1: make docstring panda-specific

src/eva/vision/data/datasets/classification/panda.py
nkaenzig: good point. I renamed it now to `_get_target_from_path`

src/eva/vision/data/datasets/classification/panda.py
nkaenzig: done

PR #440, opened 2024-05-14
Improve the mask loading in `TotalSegmentator2D`

### LINKED ISSUE #417: Improve the mask loading in `TotalSegmentator2D`
Improve the loading time of mask loading by saving and loading them as numpy arrays

ioangatop:
Closes #417


src/eva/vision/data/datasets/segmentation/total_segmentator.py
```diff
@@ -148,23 +160,54 @@ def __len__(self) -> int:
...
+        total_samples = len(self._samples_dirs)
+        for sample_index in tqdm.trange(
+            total_samples, desc=">> Exporting optimized semantic masks"
+        ):
+            masks_dir = self._get_masks_dir(sample_index)
+            filename = os.path.join(masks_dir, "semantic_labels", "masks.nii.gz")
+            if os.path.isfile(filename):
+                continue
+
+            mask_paths = [os.path.join(masks_dir, label + ".nii.gz") for label in self.classes]
```
nkaenzig, line 204: The same 4 lines of code are used above in `_load_mask `. Would it make sense to refactor those into a function?

ioangatop: right I was just doing that :D 

done

src/eva/vision/utils/io/nifti.py
```diff
@@ -28,27 +29,39 @@ def read_nifti_slice(
...
+        image_data = image_data.slicer[:, :, slice_index : slice_index + 1]  # type: ignore
+    image_array = image_data.get_fdata()  # type: ignore
     if use_storage_dtype:
         image_array = image_array.astype(image_data.get_data_dtype())  # type: ignore
     return image_array
 
 
-def fetch_total_nifti_slices(path: str) -> int:
-    """Fetches the total slides of a NIfTI image file.
+def save_array_as_nifti(array: npt.ArrayLike, filename: str) -> None:
```
nkaenzig, line 40: When should we use `np.ndarray` vs `npt.ArrayLike` as type hints? (asking because in many other places we use `np.ndarray`)

nkaenzig (approved):
LGTM

src/eva/vision/utils/io/nifti.py
ioangatop: the correct one - according to `pyright` - is the `npt.ArrayLike`

its interesting that in this case, it complained and worked only with `npt.ArrayLike`. So ideally I would say we should use the `npt.ArrayLike` everywhere - actually we should open an issue about that

src/eva/vision/data/datasets/segmentation/total_segmentator.py
```diff
@@ -70,6 +73,8 @@ def __init__(
             as_uint8: Whether to convert and return the images as a 8-bit.
             classes: Whether to configure the dataset with a subset of classes.
                 If `None`, it will use all of them.
+            optimize_mask_loading: Whether to pre-process the segmentation masks
```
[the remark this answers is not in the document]
ioangatop: great suggestion; done

src/eva/vision/utils/io/nifti.py
nkaenzig: yeah let's open an issue for this :)

roman807 (approved):

PR #443, opened 2024-05-15
Support int16 training on `TotalSegementator2D`

### LINKED ISSUE #442: Allow to train with `int16` data
Radiology images are formed in `int16` format, and covering to `uint8` we potentially loosing important details

ioangatop:
Closes #442

### DIFF
--- a/src/eva/vision/callbacks/loggers/batch/segmentation.py
+++ b/src/eva/vision/callbacks/loggers/batch/segmentation.py
@@ -68,7 +68,7 @@ class SemanticSegmentationLogger(base.BatchLogger):
         images, targets, predictions = to_cpu([images, targets, predictions])
         predictions = torch.argmax(predictions, dim=1)
 
-        images = list(map(self._denormalize_image, images))
+        images = list(map(self._format_image, images))
         targets = list(map(_draw_semantic_mask, targets))
         predictions = list(map(_draw_semantic_mask, predictions))
         image_grid = _make_grid_from_image_groups(
@@ -82,9 +82,10 @@ class SemanticSegmentationLogger(base.BatchLogger):
             step=trainer.global_step,
         )
 
-    def _denormalize_image(self, image: torch.Tensor) -> torch.Tensor:
-        """De-normalizes an image tensor to (0., 1.) range."""
-        return _denormalize_image(image, mean=self._mean, std=self._std)
+    def _format_image(self, image: torch.Tensor) -> torch.Tensor:
+        """Descaled an image tensor to (0, 255) uint8 tensor."""
+        norm_image = _descale_image(image, mean=self._mean, std=self._std)
+        return _denorm_image(norm_image)
 
 
 def _subsample_tensors(
@@ -106,8 +107,8 @@ def _subsample_tensors(
     return tensors_stack
 
 
-def _denormalize_image(
-    tensor: torch.Tensor,
+def _descale_image(
+    image: torch.Tensor,
     mean: Iterable[float] = (0.0, 0.0, 0.0),
     std: Iterable[float] = (1.0, 1.0, 1.0),
     inplace: bool = True,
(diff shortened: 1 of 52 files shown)


src/eva/vision/callbacks/loggers/batch/segmentation.py
```diff
@@ -106,34 +107,52 @@ def _subsample_tensors(
...
 
     return functional.normalize(
-        tensor,
+        image,
         mean=[-cmean / cstd for cmean, cstd in zip(mean, std, strict=False)],
         std=[1 / cstd for cstd in std],
     )
 
 
+def _denorm_image(image: torch.Tensor, inplace: bool = True) -> torch.Tensor:
```
nkaenzig, line 137: Should we move this to a public vision utils module. There might be other usecases for such functions coming up.

src/eva/vision/utils/convert.py
```diff
@@ -6,19 +6,16 @@
...
+        The image as uint8 (0, 255) array.
     """
-    if np.issubdtype(image_array.dtype, np.integer):
-        image_array = image_array.astype(np.float64)
-
-    image_scaled_array = image_array - image_array.min()
-    image_scaled_array /= image_scaled_array.max()
-    image_scaled_array *= 255
-    return image_scaled_array.astype(np.uint8)
+    image_scaled = image - image.min()
```
nkaenzig, line 18: This logic seems similar to `_denorm_image` function?

ioangatop: its exactly the same yes, but this one is for numpy arrays and the other for tensor - i didn't find a way to combine them without covering an image two times (tensor -> array -> tensor) 🤔

src/eva/vision/utils/convert.py
```diff
@@ -6,19 +6,16 @@
...
+    image_scaled /= image_scaled.max()
```
nkaenzig, line 19: Shouldn't this be `image_scaled /= (image_scaled.max() - image_scaled.min())`?

src/eva/vision/utils/convert.py
nkaenzig: Good point. At least for the min-max scaling, something like this works as torch & numpy share the same API for min/max:
```
def min_max_scale(image: npt.NDArray[Any] | torch.Tensor]) -> npt.NDArray[Any] | torch.Tensor:
    return (image - image.min()) / (image.max() - image.min())
```

The issue with that of course would be that the return type isn't uniquely defined. Could be solved maybe with something like `@functools.singledispatch`. But then going there is probably overkill for this case..
Can also leave it as is.

src/eva/vision/utils/convert.py
ioangatop: why is that? because it was already shifted from the substraction of the image min

src/eva/vision/utils/convert.py
ioangatop: actually, after conversation with radiology lab, I removed completely the numpy version, as the 8-bit range is not interested for them

So I moved the tensor implementation to public `conver` file

nkaenzig: perfect, thanks!

src/eva/vision/utils/convert.py
ioangatop: here it uses less extra variables as first the minimum value is moved to zero, thereafter one can take the maximum value to get the full range (max-min) of the specific image.

the following are the same:
```
data_min = np.min(data)
data_max = np.max(data)
scaled_data = (data - data_min) / (data_max - data_min)
```

and
```
data -= data.min()
data /= data.max()
```

here: https://stackoverflow.com/a/70783585

nkaenzig: Ah got you, makes sense!

nkaenzig (approved):

PR #446, opened 2024-05-15
Move metrics to CPU when using single device

### LINKED ISSUE #445: Add device management for the metrics calculation
The evaluation metrics can be computed both in cpu and gpu. However, when single device is used, is much more efficient to store and calculate the metrics on CPU. So we would like to have the option to select the metrics calculation device.

ioangatop:
Closes #445

### DIFF
--- a/src/eva/core/models/modules/module.py
+++ b/src/eva/core/models/modules/module.py
@@ -49,7 +49,7 @@ class ModelModule(pl.LightningModule):
 
     @property
     def metrics_device(self) -> torch.device:
-        """Returns the device which the metrics should be calculated.
+        """Returns the device by which the metrics should be calculated.
 
         We allocate the metrics to CPU when operating on single device, as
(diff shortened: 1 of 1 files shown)


src/eva/core/models/modules/module.py
```diff
@@ -46,6 +47,21 @@ def default_postprocess(self) -> batch_postprocess.BatchPostProcess:
         """The default post-processes."""
         return batch_postprocess.BatchPostProcess()
 
+    @property
+    def metrics_device(self) -> torch.device:
+        """Returns the device which the metrics should be calculated.
```
roman807, line 52: ...device **by** which...

roman807 (approved):

src/eva/core/models/modules/module.py
ioangatop: done!

PR #450, opened 2024-05-16
Remove total segmentator classification dataset

### LINKED ISSUE #449: Remove the `TotalSegmentator` classification dataset

ioangatop:
Closes #449


src/eva/vision/data/datasets/__init__.py
nkaenzig, line 1: We should also remove it from the docs:
https://github.com/kaiko-ai/eva/blob/main/docs/reference/vision/data/datasets.md?plain=1#L9

nkaenzig (approved):

src/eva/vision/data/datasets/__init__.py
ioangatop: ah good one, removed!

PR #456, opened 2024-05-17
Update WSI foreground segmentation algorithm

### LINKED ISSUE #455: Improve segmentation algorithm for WSI foreground
Currently we use a very simplistic grayscale thresholding algorithm, which doesn't work well for slides that don't have clear and uniform white background:

Two examples from camelyon16:
image

nkaenzig:
Closes #455

I used a simplified version of the segmentation algorithm from [CLAM](https://github.com/mahmoodlab/CLAM). CLAM on top of what is implemented here filters the approximate contours of the detected foreground objects based on an area threshold as a post processing step. We don't need this step here, because our patching logic anyways applies another foreground threshold filter to the sampled patches, so very small foreground areas in the mask most likely won't yield patches anyways. 

## Comparison
- "grayscale-threshold" is what we've been using so far (for datasets such as panda which have a clean white background this worked reasonably well, but it breaks for datasets such as camelyon16 where this is not the case)
- mb stands for median blur applied to saturation channel in HSV image
- otsu thresholding applied to saturation channel doesn't seem to perform better than using a fixed threshold
- I would recommend disabling the `fill_holes` in our default setup, as filled holes can result in patches sampled from within holes which might not contain relevant tissue data.

image

### Avg Runtime per WSI [s]
grayscale_threshold    0.089027
hsv              0.071794
hsv + mb            0.138696
hsv + mb + holes         0.140928
hsv + mb + otsu               0.138782
hsv + mb + otsu + holes   0.143519

- median blur is computationally the most expensive step, but it generates cleaner and smoother masks. 138ms per WSI is still fast enough, so I recommend to enable it.

### DIFF
--- a/src/eva/vision/data/wsi/__init__.py
+++ b/src/eva/vision/data/wsi/__init__.py
@@ -2,5 +2,15 @@
 
 from eva.vision.data.wsi.backends import Wsi, get_cached_wsi, wsi_backend
 from eva.vision.data.wsi.patching.coordinates import PatchCoordinates, get_cached_coords
+from eva.vision.data.wsi.patching.mask import Mask, get_mask, get_mask_level
 
-__all__ = ["Wsi", "PatchCoordinates", "get_cached_coords", "wsi_backend", "get_cached_wsi"]
+__all__ = [
+    "Wsi",
+    "PatchCoordinates",
+    "Mask",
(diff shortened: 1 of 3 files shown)


ioangatop (approved):
Looks great 🚀

PR #457, opened 2024-05-17
Normalisations and transforms for `int16` image types

### LINKED ISSUE #451: Rescale intensity of `int16` data in CT scans
We should clip the values of ct-scans from -1024 to 1024 and maybe applied windowing transform

See https://stackoverflow.com/a/72562081

ioangatop:
Closes #451


src/eva/vision/data/transforms/normalization/clamp.py
```diff
@@ -0,0 +1,43 @@
...
+            out_range: The lower and upper bound of the range to
+                be clamped to.
+        """
+        super().__init__()
+
+        self._out_range = out_range
+
+    @functools.singledispatchmethod
+    @override
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
```
nkaenzig, line 28: Nit: Would it make sense to use more specific type hints for the return type?

src/eva/vision/data/transforms/normalization/rescale_intensity.py
```diff
@@ -0,0 +1,53 @@
...
+    @_transform.register(torch.Tensor)
+    def _(self, inpt: torch.Tensor, params: Dict[str, Any]) -> Any:
+        return functional.rescale_intensity(
+            inpt, in_range=self._in_range, out_range=self._out_range
+        )
+
+    @_transform.register(tv_tensors.Image)
+    def _(self, inpt: tv_tensors.Image, params: Dict[str, Any]) -> Any:
+        scaled_inpt = functional.rescale_intensity(inpt, out_range=self._out_range)
+        return tv_tensors.wrap(scaled_inpt, like=inpt)
```
nkaenzig, line 48: Why is `.wrap` needed here? What is the benefit of wrap over `tv_tensors.Image(scaled_inp)`?

nkaenzig (commented):
Looks very good! Should we add some unit tests for the transforms and the functional modules?

src/eva/vision/data/transforms/normalization/clamp.py
roman807: maybe also `inpt` ->  `input`

ioangatop: got it from the `torchvision` source code - to be in line with them. `data` is maybe a good alternative but it depends how close we want to follow torchvision

@roman807 we can not use `input` because is a python key word

src/eva/vision/data/transforms/normalization/rescale_intensity.py
ioangatop: `wrap` is recommended because I believe it will not copy the tensor to Image, it will just cast it

ioangatop:
> Looks very good! Should we add some unit tests for the transforms and the functional modules?

Tests added 🙌

nkaenzig (approved):
LGTM

PR #461, opened 2024-05-21
Refactor embeddings writer

### LINKED ISSUE #460: Refactor `EmbeddingsWriter` class
Currently `EmbeddingsWriter` can be used only for classification tasks. However, we would like to make it as a base class, and then sub-classes like `ClassificationEmbeddingsWriter` or `SegmentationEmbeddingsWriter` to be for specific tasks.

ioangatop:
Closes #460


src/eva/core/callbacks/writers/embeddings/classification.py
```diff
@@ -0,0 +1,67 @@
...
+
+import torch
+from torch import multiprocessing
+from typing_extensions import override
+
+from eva.core.callbacks.writers.embeddings import base
+from eva.core.callbacks.writers.embeddings.typings import QUEUE_ITEM
+
+
+class ClassificationEmbeddingsWriter(base.EmbeddingsWriter):
```
nkaenzig, line 16: While the refactoring looks very nice, this seems to undo the changes added in this [PR](https://github.com/kaiko-ai/eva/pull/384), which is needed for WSI level tasks where we need to store all embeddings corresponding to the same slide to a single file?

nkaenzig: Maybe for that usecase we can implement a `ClassificationMultiEmbeddingsWriter` class or similar?

nkaenzig: Maybe we can do a `ClassificationMultiEmbeddingsWriter` or similar for that usecase though?

ioangatop: hmm I see 🤔 I think `ClassificationMultiEmbeddingsWriter` is good

but also, feel free to commit any modify / add anything that can help with the WSI level or any more refactoring ideas

nkaenzig (approved):
LGTM

PR #466, opened 2024-05-23
Add configuration logger

### LINKED ISSUE #448: Add configuration logger
Allow to save the configuration to the experimental logger

ioangatop:
Closes #448

It saves the config file in the output folder and in the loggers:
Screenshot 2024-05-23 at 3 08 07 PM

and optionally can be viewed in the terminal:
Screenshot 2024-05-23 at 3 19 18 PM


src/eva/core/callbacks/config.py
```diff
@@ -0,0 +1,169 @@
...
+    Returns:
+        The merged configurations as a dictionary.
+    """
+    merged_config = {}
+    for config_path in paths:
+        fs = cloud_io.get_filesystem(config_path)
+        with fs.open(config_path, "r") as file:
+            omegaconf_file = OmegaConf.load(file)  # type: ignore
+            config_dict = OmegaConf.to_object(omegaconf_file)  # type: ignore
+            parsed_config = _type_resolver(config_dict)  # type: ignore
```
nkaenzig, line 122: Why using a custom resolve method here and not `OmegaConf.resolve()`?

src/eva/core/callbacks/config.py
```diff
@@ -0,0 +1,169 @@
...
+        The mapping with the formatted values.
+    """
+    for key, value in mapping.items():
+        if isinstance(value, dict):
+            formatted_value = _type_resolver(value)
+        elif isinstance(value, list) and isinstance(value[0], dict):
+            formatted_value = [_type_resolver(subvalue) for subvalue in value]
+        else:
+            try:
+                parsed_value = eval(value)  # type: ignore # nosec
```
nkaenzig, line 143: Do we need `eval` here? This might be problematic from a security perspective.

nkaenzig (commented):
Looks nice! Left some minor comments.

src/eva/core/callbacks/config.py
ioangatop: `OmegaConf.resolve` doen't work in the way that we want - the `to_object` fills the environmental variables with the actual values and the custom `_type_resolver` actually parses all the strings that can be parsed, so it would do '256' -> 256, '[0.485, 0.456, 0.406]' -> [0.485, 0.456, 0.406] etc

src/eva/core/callbacks/config.py
ioangatop: `eval` is used to turn str to objects (like `'True'` -> `True` etc) and is the only one that I found that worked reliably. An alternative is to use `ast.literal_eval`, but it turns `True` -> `"yes"` 🤕 
I think its fine as the value are values from the config either way, and if it detects a buildin method / function, it will use it as a string (the next line)

src/eva/core/callbacks/config.py
nkaenzig: Do you know if `OmegaConf` has it on their roadmap to include this feature? If not, should we maybe create an issue on their repo?

src/eva/core/callbacks/config.py
ioangatop: I get staff like the following 🤔 
Screenshot 2024-05-27 at 1 21 21 PM

ioangatop: ah I actually made it work so `ats` 💪

src/eva/core/callbacks/config.py
ioangatop: we could def try, but I'm not sure if they are interested or if they are active enough

nkaenzig (approved):

PR #468, opened 2024-05-24
Add `TiffSlide` backend

### LINKED ISSUE #467: Add `TiffSlide` backend
TiffSlide is faster than OpenSlide, so for tiff based WSI datasets we might want to use it to speed up the patch embedding generation step.

nkaenzig:
Closes #467


roman807 (approved):
LGTM, thanks!

Issue #475, opened 2024-05-28
Define slide-level evaluation protocol


roman807:

roman807:
Following findings with experiments with three checkpoints of different FM sizes on Camelyon16 + Panda:
 - AdamW converges, SGD does not
 - LR=1e-3 ideal (lower doesn't reach optimal performance, takes much longer to run)
 - Number of steps 12,500: works well for large dataset (Panda 10k slides), is too many for small datasets (Camelyon 400 slides) -> cap for max n epochs necessary
 - Patience 10% is sufficient, min of 8 epochs needed
 - Batch size: 32 works as well or better than 16, 64 and bettern than >= 128

 --> Proposed protocol:
 - Number of steps: 12,500 (max: 100 epochs)
 - Patience: 10% of epochs (min 8 epochs)
 - Batch size: 32
 - LR: 1e-3
 - Optimizer: AdamW

PR #476, opened 2024-05-28
459 add camelyon16 slide level task

roman807:
closes: https://github.com/kaiko-ai/eva/issues/459

Note: config-setup is not final, will be addressed here: https://github.com/kaiko-ai/eva/issues/475


docs/datasets/camelyon16.md
```diff
@@ -0,0 +1,68 @@
+# Camelyon16
+
+The Camelyon16 dataset consists of 400 WSIs of lymph nodes for breast cancer metastasis classification. The dataset is a combination of data from two independent datasets, collected from two separate medical centers in the Netherlands (Radboud University Medical Center and University Medical Center Utrecht). The dataset contains the slides from which [PatchCamelyon](patch_camelyon.md)-patches were extracted.
```
nkaenzig, line 3: data from two independent datasets -> two independent datasets

docs/datasets/camelyon16.md
```diff
@@ -0,0 +1,68 @@
...
+│   ├── evaluation     # masks not in use
+│   ├── reference.csv  # targets
+│   └── lesion_annotations.zip
+```
+
+## Download and preprocessing
+
+The `Camelyon16` dataset class doesn't download the data during runtime and must be downloaded manually from links provided [here](https://camelyon17.grand-challenge.org/Data/).
+
+The dataset is split into train / test. As in [1], we additionally split the train set into a stratified 90:10 train/val split.
```
nkaenzig, line 54: Should we mention here that we used the same val split as in patch camelyon?

docs/datasets/camelyon16.md
```diff
@@ -0,0 +1,68 @@
...
+
+| Splits   | Train       | Validation  | Test       |  
+|----------|-------------|-------------|------------|
+| #Samples | 243 (60.75%)  | 27 (6.75%)  | 130 (32.5%) |
```
nkaenzig, line 58: Do you think the validation split might be too small?

docs/datasets/index.md
```diff
@@ -7,9 +7,10 @@
 ### Whole Slide (WSI) and microscopy image datasets
 
 #### Slide-level
-| Dataset                            | #Slides  | Slide Size        | Magnification (μm/px)  | Task                       | Cancer Type      |
-|------------------------------------|----------|------------|------------------------|----------------------------|------------------|
-| [PANDA](panda.md)                  | 3,152    | ~20k x 20k x 3    |  20x (0.5)             | Classification (6 classes) | Prostate         |
+| Dataset                            | #Slides  | Slide Size                | Magnification (μm/px)  | Task                       | Cancer Type      |
+|------------------------------------|----------|---------------------------|------------------------|----------------------------|------------------|
+| [Camelyon16](camelyon16.md)        | 400      | ~100-250k x ~100-250k x 3 |  40x (0.25)            | Classification (2 classes) | Breast           |
+| [PANDA](panda.md)                  | 3,152    | ~20k x 20k x 3            |  20x (0.5)             | Classification (6 classes) | Prostate         |
```
nkaenzig, line 13: The number of slides for PANDA seems wrong.

src/eva/vision/data/datasets/classification/camelyon16.py
```diff
@@ -0,0 +1,224 @@
...
+            sampler=sampler,
+            target_mpp=target_mpp,
+            backend=backend,
+            image_transforms=image_transforms,
+        )
+
+    @property
+    @override
+    def classes(self) -> List[str]:
+        return ["0", "1"]
```
nkaenzig, line 126: E.g. in `BACH` we use human interpretable strings in the `classes` property, together with a `class_to_idx` property.

src/eva/vision/data/datasets/classification/camelyon16.py
```diff
@@ -0,0 +1,224 @@
...
+        """Loads the dataset labels."""
+        path = os.path.join(self._root, "testing/reference.csv")
+        reference_df = pd.read_csv(path, header=None)
+        return {k: v.lower() for k, v in reference_df[[0, 1]].itertuples(index=False)}
+
+    @override
+    def prepare_data(self) -> None:
+        _validators.check_dataset_exists(self._root, True)
+
+        if not os.path.isdir(os.path.join(self._root, "training/normal")):
```
nkaenzig, line 139: Maybe we could refactor those if statements and replace by a loop?

```python
expected_resources = ["training/normal", "training/tumor", ...]

for resource in expected_resources:
    if not os.path.isdir(os.path.join(self._root, resource)):
        raise ...

```

src/eva/vision/data/datasets/classification/camelyon16.py
```diff
@@ -0,0 +1,224 @@
...
+
+        val_images_paths = [
+            path
+            for path in train_images_paths_all
+            if self._get_id_from_path(path) in self._val_slides
+        ]
+        train_images_paths = [
+            path for path in train_images_paths_all if path not in val_images_paths
+        ]
+        test_file_paths = sorted(glob.glob(os.path.join(self._root, "testing/images", "*.tif")))
```
nkaenzig, line 196:
```suggestion        
        train_paths, val_paths = [], []
        for path in glob.glob(os.path.join(self._root, "training/**/*.tif")):
            if self._get_id_from_path(path) in self._val_slides:
                val_paths.append(path)
            else:
                train_paths.append(path)
```

src/eva/vision/data/datasets/classification/camelyon16.py
```diff
@@ -0,0 +1,224 @@
...
+            image_transforms=image_transforms,
+        )
+
+    @property
+    @override
+    def classes(self) -> List[str]:
+        return ["0", "1"]
+
+    @functools.cached_property
+    def test_annotations(self) -> Dict[str, str]:
```
nkaenzig, line 129: Maybe we could rename this to `annotations` and store the annotations for train / val here as well?
Could use `self. file_paths` & `_get_target_from_path` for train/val, and the `testing/reference.csv` for loading the test labels.

src/eva/vision/data/datasets/classification/camelyon16.py
```diff
@@ -0,0 +1,224 @@
...
+            raise FileNotFoundError("'training/normal' directory not found in the root folder.")
+        if not os.path.isdir(os.path.join(self._root, "training/tumor")):
+            raise FileNotFoundError("'training/tumor' directory not found in the root folder.")
+        if not os.path.isdir(os.path.join(self._root, "testing/images")):
+            raise FileNotFoundError("'testing/images' directory not found in the root folder.")
+        if not os.path.isfile(os.path.join(self._root, "testing/reference.csv")):
+            raise FileNotFoundError("'reference.csv' file not found in the testing folder.")
+
+    @override
+    def validate(self) -> None:
```
nkaenzig, line 149: Might be good to add a check somewhere if the number of `.tif` files found matches the expected dataset size, just to make sure the user has downloaded the complete dataset.

nkaenzig (commented):
Looks good! I left some minor comments & code refactoring suggestions.

configs/vision/dino_vit/offline/camelyon16.yaml
nkaenzig, line 1: Should we add this to `test_configuration_initialization` in `tests/eva/vision/test_vision_cli.py`?

docs/datasets/camelyon16.md
roman807: yes, this is also outdated. Using the pcam split its 216 train / 54 val

src/eva/vision/data/datasets/classification/camelyon16.py
roman807: i wouldn't do this here, because `testing/reference.csv` is a raw metadata file. Better to use it original and as intended than to modify (and confuse users).

src/eva/vision/data/datasets/classification/camelyon16.py
roman807: added this, checking the number of file paths now

src/eva/vision/data/datasets/classification/camelyon16.py
roman807: added this

src/eva/vision/data/datasets/classification/camelyon16.py
roman807: changed this (kept `test_paths`)

src/eva/vision/data/datasets/classification/camelyon16.py
```diff
@@ -0,0 +1,228 @@
...
+
+        expected_n_files = {
+            "train": 216,
+            "val": 54,
+            "test": 129,
+            None: 399,
+        }
+        length = expected_n_files.get(self._split, 0)
+        _validators.check_number_of_files(self._file_paths, length, self._split)
+
```
nkaenzig, line 163: we could remove the blank line here

src/eva/vision/data/datasets/classification/camelyon16.py
```diff
@@ -0,0 +1,228 @@
...
+    def _load_file_paths(self, split: Literal["train", "val", "test"] | None = None) -> List[str]:
+        """Loads the file paths of the corresponding dataset split."""
+        train_paths, val_paths = [], []
+        for path in glob.glob(os.path.join(self._root, "training/**/*.tif")):
+            if self._get_id_from_path(path) in self._val_slides:
+                val_paths.append(path)
+            else:
+                train_paths.append(path)
+        test_paths = sorted(glob.glob(os.path.join(self._root, "testing/images", "*.tif")))
```
nkaenzig, line 200: I think we should apply the `sorted` only in line 213 to all the paths that are returned. Currently the sort is only applied to the test paths, so the order of the train / val paths is not deterministic

src/eva/vision/data/datasets/classification/camelyon16.py
nkaenzig: I'm not suggesting to modify the `testing/reference.csv` file, but to have a `annotations` property that contains all annotations, and not only the ones of the test set. 

Two me this has two main advantages:
1. Cleaner code structure
2. Simplifies the `_get_target_from_path` function to:
```python
def _get_target_from_path(self, file_path: str) -> int:
    return self.annotations_test_set[self._get_id_from_path(file_path)]
```

Probably then we don't even need a seperate function for `_get_target_from_path`, but we could do that directly in the `load_target` function.

src/eva/vision/data/datasets/classification/camelyon16.py
```diff
@@ -0,0 +1,228 @@
...
+            case _:
+                raise ValueError("Invalid split. Use 'train', 'val' or `None`.")
+        return [os.path.relpath(path, self._root) for path in paths]
+
+    def _get_target_from_path(self, file_path: str) -> int:
+        """Returns the target label based on the file path."""
+        slide_id = self._get_id_from_path(file_path)
+        test_annotations = self.annotations_test_set.get(slide_id, "")
+
+        for class_name in self.classes:
```
nkaenzig, line 220: Would be good if we could get rid of the for loop here, the proposal [here](https://github.com/kaiko-ai/eva/pull/476#discussion_r1626074067) would be one option to do so.

src/eva/vision/data/datasets/classification/camelyon16.py
```diff
@@ -0,0 +1,228 @@
...
+    @override
+    def validate(self) -> None:
+
+        expected_n_files = {
+            "train": 216,
+            "val": 54,
+            "test": 129,
+            None: 399,
+        }
+        length = expected_n_files.get(self._split, 0)
```
nkaenzig, line 161: nit: I would use `expected_n_files[self._split]` and let it fail if the split has an unexpected value, rather than falling back to 0

src/eva/vision/data/datasets/classification/camelyon16.py
```diff
@@ -0,0 +1,228 @@
...
+            if not os.path.isdir(os.path.join(self._root, resource)):
+                raise FileNotFoundError(f"'{resource}' not found in the root folder.")
+
+        if not os.path.isfile(os.path.join(self._root, "testing/reference.csv")):
+            raise FileNotFoundError("'reference.csv' file not found in the testing folder.")
+
+    @override
+    def validate(self) -> None:
+
+        expected_n_files = {
```
nkaenzig, line 155: How did you deal with the issue that in unit tests we use a smaller subset of images? For panda I had to do some hacks / patching to make it work, but here I can't find any logic related to that?

src/eva/vision/data/datasets/classification/camelyon16.py
roman807: refactored this -- packed the logic into `annotations` and got rid off `_get_target_from_path`. Let me know if this is how you think it should look like?

src/eva/vision/data/datasets/classification/camelyon16.py
roman807: currently `valdate()` is not called for MultiWsiDatasets --> created an issue for this: https://github.com/kaiko-ai/eva/issues/499

src/eva/vision/data/datasets/classification/camelyon16.py
```diff
@@ -0,0 +1,244 @@
...
+            reference_df = pd.read_csv(path, header=None)
+            annotations.update(
+                {k: v.lower() for k, v in reference_df[[0, 1]].itertuples(index=False)}
+            )
+
+        if self._split in ["train", "val", None]:
+            annotations.update(
+                {
+                    self._get_id_from_path(file_path): self._get_class_from_path(file_path)
+                    for file_path in self._file_paths
```
nkaenzig, line 155: In the `None` case `self._file_paths` also contains the test paths, so then you would overwrite the test labels if I see correctly?

src/eva/vision/data/datasets/classification/camelyon16.py
nkaenzig: amazing! thanks for the refactoring

src/eva/vision/data/datasets/classification/camelyon16.py
roman807: updated this (previously it would not have worked for None, since `_get_class_from_path` only works with train file pahts

nkaenzig (approved):
LGTM

PR #478, opened 2024-05-28
Update the vision dataset return types to `tv_tensors`

### LINKED ISSUE #477: Update the dataset return types to `tv_tensors`
Currently, from the vision datasets we return types like `torch.Tensor` or `numpy` arrays.

We would like to update and return torchvision tensors (`tv_tensors`) in order to be native compatible with torchvision transforms v2 and better handling of specific tensor types (use `tv_tensors.Image`, `tv_tensors.Mask`, `tv_tensors.BBox` instead annotate them all with `torch.Tensor`)

ioangatop:
Closes #477

### DIFF
--- a/src/eva/vision/data/datasets/classification/mhist.py
+++ b/src/eva/vision/data/datasets/classification/mhist.py
@@ -3,7 +3,8 @@
 import os
 from typing import Callable, Dict, List, Literal, Tuple
 
-import numpy as np
+import torch
+from torchvision import tv_tensors
 from typing_extensions import override
 
 from eva.vision.data.datasets import _validators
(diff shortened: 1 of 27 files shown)


src/eva/vision/data/datasets/classification/mhist.py
```diff
@@ -74,16 +69,16 @@ def validate(self) -> None:
...
-        return io.read_image(image_path)
+        return io.read_image_as_tensor(image_path)
 
     @override
-    def load_target(self, index: int) -> np.ndarray:
+    def load_target(self, index: int) -> torch.Tensor:
         _, label = self._samples[index]
         target = self.class_to_idx[label]
-        return np.asarray(target, dtype=np.int64)
+        return torch.tensor(target, dtype=torch.float32)
```
nkaenzig, line 81: In CRC we use `torch.long` and here `torch.float32`. What is the reason for that?

ioangatop: its for binary classification (mhist and pcam), the criterion need a float from 0 - 1; currently we transform the type to float during transforms but I figured since we will do it either-way maybe we can do straight from the dataset

nkaenzig (approved):
LGTM

PR #511, opened 2024-06-07
475 define slide level evaluation protocol

roman807:
closes https://github.com/kaiko-ai/eva/issues/475 (see protocol details in issue)

### DIFF
--- a/configs/vision/dino_vit/offline/camelyon16.yaml
+++ b/configs/vision/dino_vit/offline/camelyon16.yaml
@@ -131,4 +131,3 @@ data:
         batch_size: *BATCH_SIZE
       predict:
         batch_size: &PREDICT_BATCH_SIZE ${oc.env:PREDICT_BATCH_SIZE, 64}
-        prefetch_factor: 2
--- a/configs/vision/dino_vit/offline/camelyon16.yaml
+++ b/configs/vision/dino_vit/offline/camelyon16.yaml
@@ -24,7 +24,7 @@ trainer:
           mode: *MONITOR_METRIC_MODE
       - class_path: eva.callbacks.EmbeddingsWriter
         init_args:
-          output_dir: &DATASET_EMBEDDINGS_ROOT ${oc.env:EMBEDDINGS_ROOT, ./data/embeddings/dino_vits16/camelyon16}
+          output_dir: &DATASET_EMBEDDINGS_ROOT ${oc.env:EMBEDDINGS_ROOT, ./data/embeddings/${oc.env:DINO_BACKBONE, dino_vits16}/camelyon16}
           dataloader_idx_map:
             0: train
             1: val
@@ -131,5 +131,4 @@ data:
         batch_size: *BATCH_SIZE
       predict:
         batch_size: &PREDICT_BATCH_SIZE ${oc.env:PREDICT_BATCH_SIZE, 64}
-        num_workers: ${oc.env:NUM_WORKERS, 12}
         prefetch_factor: 2
--- a/configs/vision/dino_vit/offline/panda.yaml
+++ b/configs/vision/dino_vit/offline/panda.yaml
@@ -131,4 +131,3 @@ data:
         batch_size: *BATCH_SIZE
       predict:
         batch_size: &PREDICT_BATCH_SIZE ${oc.env:PREDICT_BATCH_SIZE, 64}
-        prefetch_factor: 2
--- a/configs/vision/dino_vit/offline/panda.yaml
+++ b/configs/vision/dino_vit/offline/panda.yaml
@@ -24,7 +24,7 @@ trainer:
           mode: *MONITOR_METRIC_MODE
       - class_path: eva.callbacks.EmbeddingsWriter
         init_args:
-          output_dir: &DATASET_EMBEDDINGS_ROOT ${oc.env:EMBEDDINGS_ROOT, ./data/embeddings}
+          output_dir: &DATASET_EMBEDDINGS_ROOT ${oc.env:EMBEDDINGS_ROOT, ./data/embeddings/${oc.env:DINO_BACKBONE, dino_vits16}/panda}
           dataloader_idx_map:
             0: train
             1: val
@@ -131,5 +131,4 @@ data:
         batch_size: *BATCH_SIZE
       predict:
         batch_size: &PREDICT_BATCH_SIZE ${oc.env:PREDICT_BATCH_SIZE, 64}
-        num_workers: ${oc.env:NUM_WORKERS, 12}
         prefetch_factor: 2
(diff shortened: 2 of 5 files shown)


configs/vision/dino_vit/offline/camelyon16.yaml
```diff
@@ -19,12 +19,12 @@ trainer:
       - class_path: lightning.pytorch.callbacks.EarlyStopping
         init_args:
           min_delta: 0
-          patience: 74
+          patience: ${oc.env:PATIENCE, 10}
           monitor: *MONITOR_METRIC
           mode: *MONITOR_METRIC_MODE
       - class_path: eva.callbacks.EmbeddingsWriter
         init_args:
-          output_dir: &DATASET_EMBEDDINGS_ROOT ${oc.env:EMBEDDINGS_ROOT, ./data/embeddings}/${oc.env:DINO_BACKBONE, dino_vits16}/camelyon16
```
ioangatop, line 27: I think here is good to bring back the `/${oc.env:DINO_BACKBONE, dino_vits16}` so to save it to a different folder

configs/vision/dino_vit/offline/camelyon16.yaml
```diff
@@ -123,13 +123,13 @@ data:
...
+        batch_size: &BATCH_SIZE ${oc.env:BATCH_SIZE, 32}
         shuffle: true
       val:
         batch_size: *BATCH_SIZE
       test:
         batch_size: *BATCH_SIZE
       predict:
         batch_size: &PREDICT_BATCH_SIZE ${oc.env:PREDICT_BATCH_SIZE, 64}
-        num_workers: 12 #multiprocessing.cpu_count
+        num_workers: ${oc.env:NUM_WORKERS, 12}
```
ioangatop, line 134: i dont remember why do we have to  the `multiprocessing.cpu_count` is too much?

configs/vision/owkin/phikon/offline/camelyon16.yaml
```diff
@@ -0,0 +1,131 @@
+---
```
ioangatop, line 1: we should add these configs in the tests, the `test_configuration_initialization` to test that the config can be parsed

ioangatop (approved):

configs/vision/dino_vit/offline/camelyon16.yaml
roman807: removed `num_workers` here since it uses `multiprocessing.cpu_count()` by default

configs/vision/dino_vit/offline/panda.yaml
```diff
@@ -19,12 +19,12 @@ trainer:
...
         init_args:
           min_delta: 0
-          patience: 13
+          patience: ${oc.env:PATIENCE, 8}
           monitor: *MONITOR_METRIC
           mode: *MONITOR_METRIC_MODE
       - class_path: eva.callbacks.EmbeddingsWriter
         init_args:
-          output_dir: &DATASET_EMBEDDINGS_ROOT ${oc.env:EMBEDDINGS_ROOT, ./data/embeddings}/${oc.env:DINO_BACKBONE, dino_vits16}/panda
+          output_dir: &DATASET_EMBEDDINGS_ROOT ${oc.env:EMBEDDINGS_ROOT, ./data/embeddings/${oc.env:DINO_BACKBONE, dino_vits16}/panda}
```
ioangatop, line 27: there is an extra `}` in the end by accident

configs/vision/owkin/phikon/offline/camelyon16.yaml
```diff
@@ -0,0 +1,130 @@
...
+trainer:
+  class_path: eva.Trainer
+  init_args:
+    n_runs: &N_RUNS ${oc.env:N_RUNS, 5}
+    default_root_dir: &OUTPUT_ROOT ${oc.env:OUTPUT_ROOT, logs/owkin/phikon/offline/camelyon16}
```
ioangatop, line 6: I think here we should do `&OUTPUT_ROOT ${oc.env:OUTPUT_ROOT, logs}/owkin/phikon/offline/camelyon16`

configs/vision/owkin/phikon/offline/camelyon16.yaml
```diff
@@ -0,0 +1,130 @@
...
+      train:
+        batch_size: &BATCH_SIZE ${oc.env:BATCH_SIZE, 32}
+        shuffle: true
+      val:
+        batch_size: *BATCH_SIZE
+      test:
+        batch_size: *BATCH_SIZE
+      predict:
+        batch_size: &PREDICT_BATCH_SIZE ${oc.env:PREDICT_BATCH_SIZE, 64}
+        prefetch_factor: 2
```
ioangatop, line 130: I think `prefetch_factor=2` is by default (is it?) so maybe we can remove it?

configs/vision/owkin/phikon/offline/panda.yaml
```diff
@@ -0,0 +1,129 @@
+---
+trainer:
+  class_path: eva.Trainer
+  init_args:
+    n_runs: &N_RUNS ${oc.env:N_RUNS, 5}
+    default_root_dir: &OUTPUT_ROOT ${oc.env:OUTPUT_ROOT, logs/owkin/phikon/offline/panda}
```
ioangatop, line 6: same here as above; `default_root_dir: &OUTPUT_ROOT ${oc.env:OUTPUT_ROOT, logs}/owkin/phikon/offline/panda`

configs/vision/owkin/phikon/offline/camelyon16.yaml
roman807: i thought it would be nice to give the user full control over the log dir. E.g. if you run many experiments, you might want to indicate some used settings in the folder name which makes analysis more convenient.

configs/vision/dino_vit/offline/panda.yaml
roman807: moved the parenthesis for the same reason - to give the user control over the complete path name

configs/vision/owkin/phikon/offline/camelyon16.yaml
ioangatop: that also make sense, but i though that since we have the output hash logger, it will create a unique folder either-way right? The main reason though is that if you want to run multiple experiments and save it in another destination (say `eva-logs/` instead of 'logs/') you have to reset for every config that you are doing the new path -- in the otherway, you just set the `OUTPUT_ROOT` once

however, I see that in the other configs we do the same as here, so i'm proposing lets keep it like this and if we think the other approach is better, we update all of them

configs/vision/owkin/phikon/offline/panda.yaml
roman807: (same solution as suggested in comment above)

Issue #499, opened 2024-06-06
Make sure MultiWsiDataset are validated


roman807:
Currently the `base.Dataset.setup` & `validate` ([here](https://github.com/kaiko-ai/eva/blob/2efbca5f564945e8e4fd2950db4de8f5597b83d1/src/eva/core/data/datasets/base.py)) methods are not called for MultiWsiDatasets (e.g. panda, camelyon16) as they get overwritten MultiWsiDatasets ([here](https://github.com/kaiko-ai/eva/blob/2efbca5f564945e8e4fd2950db4de8f5597b83d1/src/eva/vision/data/datasets/wsi.py))

PR #512, opened 2024-06-07
make sure multiwsidatasets are validated

roman807:
closes https://github.com/kaiko-ai/eva/issues/499

### DIFF
--- a/src/eva/core/data/datasets/base.py
+++ b/src/eva/core/data/datasets/base.py
@@ -26,10 +26,6 @@ class Dataset(dataset.TorchDataset):
         """
         self.configure()
         self.validate()
-        self._setup()
-
-    def _setup(self) -> None:
-        """Setups the datasets. This method should be overridden by the subclasses."""
 
     def configure(self):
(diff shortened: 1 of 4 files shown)


src/eva/core/data/datasets/base.py
```diff
@@ -26,6 +26,10 @@ def setup(self) -> None:
         """
         self.configure()
         self.validate()
+        self._setup()
+
+    def _setup(self) -> None:
```
ioangatop, line 31: why we have this method? for setups we should use the `self.configure()` method, or is this a blocker?

roman807: i see, @nkaenzig do you see anything that speaks against removing `_setup` and replacing it with `configure` in `MultiWsiDataset`?

nkaenzig: I think using configure is fine, I had forgotten this existed :D

roman807: move validate to configure

PR #547, opened 2024-06-25
Add `CoNSeP` dataset with patching

nkaenzig:
Closes #175 (Dataset version that applies patching)

PR #549, opened 2024-06-27
Add `MoNuSac` segmentation dataset

### LINKED ISSUE #546: Add MoNuSAC dataset
https://monusac-2020.grand-challenge.org/Data/

ioangatop:
Closes #546

image


configs/vision/dino_vit/online/monusac.yaml
nkaenzig, line 1: Please add a documentation page with the specs of the dataset.

configs/vision/dino_vit/online/monusac.yaml
```diff
@@ -0,0 +1,103 @@
...
+          # Set `download: true` to download the dataset from https://monusac-2020.grand-challenge.org/Data/
+          # The MoNuSAC dataset is distributed under the following license: 
+          # "Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International"
+          # (see: https://creativecommons.org/licenses/by-nc-sa/4.0/legalcode)
+          transforms:
+            class_path: eva.vision.data.transforms.common.ResizeAndCrop
+            init_args:
+              mean: *NORMALIZE_MEAN
+              std: *NORMALIZE_STD
+      val:
```
nkaenzig, line 93: For datasets with only 2 splits, we often have a naming issue, because when the authors call it "test", we actually use it as a "val" dataset. Same here: https://github.com/kaiko-ai/eva/pull/547

In the leaderboard, in fact we need to report the metrics as val metrics - which is confusing because they refer to the test set. I think we need to define a clear convention here on how to solve this confusion.

nkaenzig, line 93: For this dataset, would it be possible to generate our own val split, or is it too small / data leakage issues?

nkaenzig (commented):
Looks very nice. Left two minor comments.

configs/vision/dino_vit/online/monusac.yaml
ioangatop: good point -- on it!

configs/vision/dino_vit/online/monusac.yaml
ioangatop: right yeah, I think the convention is that we use the `test` split as validation, so in leaderboard we will report it as `val` -- we treat the MHIST dataset the same

ioangatop: we can avoid the data leakage, as they do provide the patient id, however the dataset is indeed small; train 209 sample and test 85

configs/vision/dino_vit/online/monusac.yaml
ioangatop: done!

docs/datasets/monusac.md
```diff
@@ -0,0 +1,66 @@
...
+The dataset class `MoNuSAC` supports downloading the data during runtime by setting the init argument `download=True`.
+
+> [!NOTE]
+> In the provided `MoNuSAC`-config files the download argument is set to `false`. To enable automatic download you will need to open the config and set `download: true`.
+
+### Splits
+
+We work with the splits provided by the data source.
+
+| Splits   | Train           | Test         |
```
roman807, line 51: if we don't have a validation set and validate on the test set, lets refer to the test set as validation set (as we did e.g. for mhist: https://kaiko-ai.github.io/eva/latest/datasets/mhist/)

roman807: also best to align the language (and remove note in line 56/57), since were describing the same issue

configs/vision/dino_vit/online/monusac.yaml
roman807: since some images exceed 224x224, would it make sense to do patching (like here: https://github.com/kaiko-ai/eva/pull/547/files#diff-d0df374ae39abf8b6d321c9334cfcfcd8d41aa0e344daa741fdb6df95f43f5b7) which would increase the dataset / potentially allow for a val split?

docs/datasets/monusac.md
```diff
@@ -0,0 +1,66 @@
...
+│   │   ├── TCGA-5P-A9K0-01Z-00-DX1_1.tif   # tissue image
+│   │   ├── TCGA-5P-A9K0-01Z-00-DX1_1.xml   # annotations
+│   │   └── ...
+```
+
+## Download and preprocessing
+The dataset class `MoNuSAC` supports downloading the data during runtime by setting the init argument `download=True`.
+
+> [!NOTE]
+> In the provided `MoNuSAC`-config files the download argument is set to `false`. To enable automatic download you will need to open the config and set `download: true`.
```
roman807, line 45: Lets align the Note style and language with those of the other datasets (if you prefer this one, change the others as well)

docs/datasets/monusac.md
```diff
@@ -0,0 +1,66 @@
+# MoNuSAC
```
roman807, line 1: add this file to `mkdocs.yaml`

src/eva/vision/data/datasets/segmentation/monusac.py
```diff
@@ -0,0 +1,239 @@
...
+                )
+                semantic_labels[fill_col_coords, fill_row_coords] = class_id
+
+        return semantic_labels
+
+    def _download_dataset(self) -> None:
+        """Downloads the dataset."""
+        self._print_license()
+        for resource in self._resources:
+            if os.path.isdir(self._root):
```
roman807, line 227: looks like this will only download the first dataset (train), and then skip the test because then `self._root` exists

ioangatop: ah good point -- fixed!

configs/vision/dino_vit/online/monusac.yaml
ioangatop: Not sure... now I do random crops so it should capture all the regions, so not sure we need it;  maybe on the offline if we dont go `CenterCrop` transform -- we can investigate in the future though

docs/datasets/monusac.md
ioangatop: ok done

docs/datasets/monusac.md
ioangatop: this is the official markdown note style, so probably nice to follow -- ok I update all of them

docs/datasets/monusac.md
ioangatop: nice done

roman807 (approved):

PR #558, opened 2024-07-04
Change `overwrite` default value to True in `EmbeddingWriter`

### LINKED ISSUE #557: `overwrite=False` in `EmbeddingsWriter` doesn't work
Setting `overwrite=False` should raise an Exception if the embeddings already exist in the specified output directory. Currently it just proceeds and overwrites the files.

nkaenzig:
Closes #557 

Enabling overwriting embeddings by default is risky, as users might unintentionally overwrite previously computed embeddings.

### DIFF
--- a/configs/vision/dino_vit/offline/bach.yaml
+++ b/configs/vision/dino_vit/offline/bach.yaml
@@ -39,6 +39,7 @@ trainer:
                 pretrained: ${oc.env:PRETRAINED, true}
                 force_reload: ${oc.env:FORCE_RELOAD, false}
               checkpoint_path: ${oc.env:CHECKPOINT_PATH, null}
+          overwrite: false
     logger:
       - class_path: lightning.pytorch.loggers.TensorBoardLogger
         init_args:
--- a/src/eva/core/callbacks/writers/embeddings/base.py
+++ b/src/eva/core/callbacks/writers/embeddings/base.py
@@ -168,18 +168,16 @@ class EmbeddingsWriter(callbacks.BasePredictionWriter, abc.ABC):
 
     def _check_if_exists(self) -> None:
         """Checks if the output directory already exists and if it should be overwritten."""
-        exists = (
-            os.path.isdir(self._output_dir) and len(list(Path(self._output_dir).rglob("*.pt"))) > 0
-        )
-        if exists and not self._overwrite:
+        try:
+            os.makedirs(self._output_dir, exist_ok=self._overwrite)
+        except FileExistsError as e:
             raise FileExistsError(
-                f"Embeddings already exists in {self._output_dir}. This either means that they "
-                "have been computed before or that a wrong output directory is being used. "
-                "Consider using `eva fit` instead, selecting a different output directory or "
-                "setting overwrite=True."
+                f"The embeddings output directory already exists: {self._output_dir}. This "
+                "either means that they have been computed before or that a wrong output "
+                "directory is being used. Consider using `eva fit` instead, selecting a "
+                "different output directory or setting overwrite=True."
             )
-        else:
-            os.makedirs(self._output_dir, exist_ok=True)
+        os.makedirs(self._output_dir, exist_ok=True)
 
 
 def _as_io_buffers(*items: torch.Tensor) -> Sequence[io.BytesIO]:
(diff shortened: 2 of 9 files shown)


configs/vision/dino_vit/offline/bach.yaml
```diff
@@ -39,6 +39,7 @@ trainer:
                 pretrained: ${oc.env:PRETRAINED, true}
                 force_reload: ${oc.env:FORCE_RELOAD, false}
               checkpoint_path: ${oc.env:CHECKPOINT_PATH, null}
+          overwrite: false
```
ioangatop, line 42: maybe lets remove the `overwrite` here, since the default should be false and to not overpopulate the config

nkaenzig: the default used to be `False`, seems that at some point this got changed (maybe by accident), which is why I now defined this explicitely

src/eva/core/callbacks/writers/embeddings/base.py
```diff
@@ -163,6 +166,21 @@ def _get_item_metadata(
...
+        )
+        if exists and not self._overwrite:
+            raise FileExistsError(
+                f"Embeddings already exists in {self._output_dir}. This either means that they "
+                "have been computed before or that a wrong output directory is being used. "
+                "Consider using `eva fit` instead, selecting a different output directory or "
+                "setting overwrite=True."
+            )
+        else:
+            os.makedirs(self._output_dir, exist_ok=True)
```
ioangatop, line 182: we can remove the `else` here, just an empty line with `os.makedirs(self._output_dir, exist_ok=True)` will do

src/eva/core/callbacks/writers/embeddings/base.py
```diff
@@ -163,6 +166,21 @@ def _get_item_metadata(
...
+        exists = (
+            os.path.isdir(self._output_dir) and len(list(Path(self._output_dir).rglob("*.pt"))) > 0
+        )
+        if exists and not self._overwrite:
+            raise FileExistsError(
+                f"Embeddings already exists in {self._output_dir}. This either means that they "
+                "have been computed before or that a wrong output directory is being used. "
+                "Consider using `eva fit` instead, selecting a different output directory or "
+                "setting overwrite=True."
+            )
```
ioangatop, line 180: this could be an alternative, but not sure if a good idea
```py3
try:
   os.makedirs(self._output_dir, exist_ok=self._overwrite)
except FileExistsError as e:
            raise FileExistsError(
                f"Embeddings already exists in {self._output_dir}. This either means that they "
                "have been computed before or that a wrong output directory is being used. "
                "Consider using `eva fit` instead, selecting a different output directory or "
                "setting overwrite=True."
            )

File: src/eva/core/callbacks/writers/embeddings/base.py [hunk 4]
```diff
@@ -163,6 +166,21 @@ def _get_item_metadata(
 
         return item_metadata
 
+    def _check_if_exists(self) -> None:
+        """Checks if the output directory already exists and if it should be overwritten."""
+        exists = (
+            os.path.isdir(self._output_dir) and len(list(Path(self._output_dir).rglob("*.pt"))) > 0
```
ioangatop, line 172: I'm worried a bit that we list all the `.pt` files because (i) what if they are many (ii) if we in the future allow to store embeddings in the cloud, this check will be challenging - maybe I'm missing something though 🤔

configs/vision/dino_vit/offline/bach.yaml
ioangatop: I see, lets keep it them :D

src/eva/core/callbacks/writers/embeddings/base.py
nkaenzig: That's a good point indeed. Solved this by replacing it with suggestion from [here](https://github.com/kaiko-ai/eva/pull/558#discussion_r1671872450)

src/eva/core/callbacks/writers/embeddings/base.py
nkaenzig: done

src/eva/core/callbacks/writers/embeddings/base.py
nkaenzig: done

ioangatop (approved):
awesome 🙏

PR #559, opened 2024-07-09
Add `BCSS` dataset

### LINKED ISSUE #553: Add `BCSS` segmentation dataset
[The BCSS dataset](https://github.com/CancerDataScience/CrowdsourcingDataset-Amgadetal2019) contains over 20,000 segmentation annotations of tissue region from breast cancer images from TCGA - https://bcsegmentation.grand-challenge.org/

Repo: https://github.com/PathologyDataScience/BCSS?tab=readme-ov-file

nkaenzig:
Closes #553 

After 4 epochs using ImageNet pre-trained `vit_small_patch16_224_dino` encoder: 
image


ioangatop (commented):
looks great, cool that we mix some minor staff as well - the only think that I have to ask of course if for the split, thoughts?

src/eva/vision/data/datasets/segmentation/bcss.py
```diff
@@ -0,0 +1,230 @@
...
+                test_indices.append(i)
+            else:
+                train_val_indices.append(i)
+
+        train_sub_indices, val_sub_indices, _ = splitting.random_split(
+            samples=train_val_indices,
+            train_ratio=self._train_split_ratio,
+            val_ratio=self._val_split_ratio,
+            seed=self._seed,
+        )
```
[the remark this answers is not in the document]
nkaenzig: Good point. in this dataset, each .png corresponds to a different TCGA slide, so at slide-level there is no data leakage. Also, all images seem to come from different patients, so also no patient-level data leakage.
However, we have multiple images from the same medical centers. The test set contains only images from centers that are not present in the train set. For the random train/val split we do here, the centers might be overlapping, so there is "leakage" at the center-level, but I'd still consider this a realistic scenario for train/val.

src/eva/vision/data/datasets/segmentation/bcss.py
```diff
@@ -0,0 +1,230 @@
...
+    """Validation split ratio."""
+
+    _expected_length: int = 151
+    """Expected dataset length."""
+
+    def __init__(
+        self,
+        root: str,
+        sampler: samplers.Sampler,
+        split: Literal["train", "val", "test"] | None = None,
```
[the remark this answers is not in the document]
nkaenzig: done

docs/datasets/bcss.md
```diff
@@ -0,0 +1,59 @@
+# BCSS
+
+The BCSS (Breast Cancer Semantic Segmentation) consists of extracts from 151 WSI images from [TCGA](https://www.cancer.gov/ccg/research/genome-sequencing/tcga), containing over 20.000 segmentation annotations covering 21 different tissue types.
```
roman807, line 3: nit: 20.000 -> 20,000

src/eva/vision/data/datasets/segmentation/bcss.py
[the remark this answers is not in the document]
nkaenzig: Haven't run a lot of experiments yet, I wanted to leave that for when we have a clearer training protocol for segmentation. So far I only validated that it learns the task.

docs/datasets/bcss.md
nkaenzig: fixed

src/eva/vision/data/datasets/segmentation/consep.py
```diff
@@ -145,23 +142,11 @@ def _load_file_paths(self, split: Literal["train", "val"] | None = None) -> List
...
         filename = self.filename(index).split(".")[0]
-        mask_dir = "Train" if filename.startswith("train") else "Test"
-        return os.path.join(self._root, mask_dir, "Labels", f"{filename}.mat")
-
-    def _extract_mask_patch(self, index: int, mask: npt.NDArray[Any]) -> npt.NDArray[Any]:
-        """Reads the mask patch at the coordinates corresponding to the given patch index."""
-        (x, y), width, height = self._get_coords(index)
-        return mask[x : x + width, y : y + height]
+        mask_dir = "Train/Labels" if filename.startswith("train") else "Test/Labels"
+        return os.path.join(self._root, mask_dir, f"{filename}.mat")
```
ioangatop, line 149: I think we shouldn't change that,  because windows used for example will have issues, as we have the `/` pre-determined here for the split

so this is safare 
```
mask_dir = "Train" if filename.startswith("train") else "Test"
return os.path.join(self._root, mask_dir, "Labels", f"{filename}.mat")
```

src/eva/vision/data/datasets/segmentation/bcss.py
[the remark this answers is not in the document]
ioangatop: make sense, we can configure it afterwards

src/eva/vision/data/datasets/segmentation/bcss.py
```diff
@@ -0,0 +1,240 @@
...
+    def __init__(
+        self,
+        root: str,
+        sampler: samplers.Sampler,
+        split: Literal["train", "val", "trainval", "test"] | None = None,
+        width: int = 224,
+        height: int = 224,
+        target_mpp: float = 0.5,
+        transforms: Callable | None = None,
+        seed: int = 42,
```
ioangatop, line 58: I would prefere to maybe remove the seed from here and fix the dataset on the 42, and we can control the overall seed of eva from only the global one, what do you think?

src/eva/vision/data/datasets/segmentation/bcss.py
```diff
@@ -0,0 +1,240 @@
...
+    _val_split_ratio: float = 0.2
+    """Validation split ratio."""
+
+    _expected_length: int = 151
+    """Expected dataset length."""
+
+    def __init__(
+        self,
+        root: str,
+        sampler: samplers.Sampler,
```
ioangatop, line 52: it comes handy to have maybe a default value here (like the grid sampler), especially when using the dataset from python. Because its a class, we can do `sampler: samplers.Sampler | None` and the init to set it to the default value if `None`. Any thoughts?

src/eva/vision/data/datasets/segmentation/bcss.py
```diff
@@ -0,0 +1,240 @@
...
+        split: Literal["train", "val", "trainval", "test"] | None = None,
+        width: int = 224,
+        height: int = 224,
```
ioangatop, line 55: maybe at a different PR but I would expect these values to be at the sampler's `__init__` values, as I believe this is major part of it (on the class side I think it makes a bit more sense to add it in the init args instead of the `sample` method) - thoughts?

src/eva/vision/data/datasets/segmentation/bcss.py
```diff
@@ -0,0 +1,240 @@
...
+        target_mpp: float = 0.5,
```
ioangatop, line 56: this `target_mpp` doesn't do much here right? make we can remove it and move it to the init hardcoded to this value

ioangatop (commented):
Looks great! here couple of comments, I think we will merge it super soon

src/eva/vision/data/datasets/segmentation/bcss.py
nkaenzig: I'll remove the seed & random split here and replace by a static `val` dataset based on a selection of centers as discussed.

src/eva/vision/data/datasets/segmentation/bcss.py
nkaenzig: good idea, done

src/eva/vision/data/datasets/segmentation/bcss.py
nkaenzig: I'd like to keep this here. In fact, having this as a flexible option helped me find during experimentation that mpp=0.25 yielded patches that were too small. Maybe users for certain FM evaluations might want to operate at different mpps, so I think it'd be nice to expose this as configurable option.

src/eva/vision/data/datasets/segmentation/bcss.py
nkaenzig: Good point. On the class side of things I agree it'd be cleaner to have it in the `sampler` init. But from a user perspective, having it exposed at the dataset init might be easier to understand / more intuitive.

No strong preference though, we can implement this in a follow up PR.

src/eva/vision/data/datasets/segmentation/bcss.py
nkaenzig: done

src/eva/vision/data/datasets/segmentation/bcss.py
nkaenzig: Opened an issue for this here: https://github.com/kaiko-ai/eva/issues/568

src/eva/vision/data/datasets/segmentation/bcss.py
nkaenzig: Probably also makes sense to refactor this together with `width` & `height` in a follow up PR: https://github.com/kaiko-ai/eva/pull/559#discussion_r1678863002

ioangatop (approved):

PR #567, opened 2024-07-12
Add `LiTS` radiology ct segmentation dataset and task

### LINKED ISSUE #564: Add LITS radiology CT segmentation dataset
https://competitions.codalab.org/competitions/17094

ioangatop:
Closes #564

TODO:
- [x] documentation 
- [x] offline workflow


src/eva/vision/data/datasets/segmentation/lits.py
```diff
@@ -0,0 +1,178 @@
...
+
+        self._root = root
+        self._split = split
+
+        self._indices: List[Tuple[int, int]] = []
+
+    @property
+    @override
+    def classes(self) -> List[str]:
+        return ["liver", "tumor"]
```
nkaenzig, line 74: In `consep.py` we list "background" as a class here as well. If not doing so, I think the mapping you get in `class_to_idx` will be wrong (the idx there should refer to the values in the target masks).

src/eva/vision/data/datasets/segmentation/lits.py
```diff
@@ -0,0 +1,178 @@
...
+        indices = [
+            (sample_idx, slide_idx)
+            for sample_idx in self._get_split_indices()
+            for slide_idx in range(self._get_number_of_slices_per_volume(sample_idx))
+            if slide_idx % (self._sample_every_n_slices or 1) == 0
+        ]
+        return indices
+
+    def _get_split_indices(self) -> List[int]:
+        """Returns the samples indices that corresponding the dataset split and version."""
```
nkaenzig, line 163: -> """Returns the sample indices for the specified dataset split."""

src/eva/core/utils/operations.py
```diff
@@ -0,0 +1,26 @@
+"""Functional operations."""
+
+import re
+from typing import Iterable, List
+
+
+def numeric_sort(item: Iterable[str], /) -> List[str]:
```
nkaenzig, line 7: what does the `, /` do here in the function args?

nkaenzig, line 7: Can you write some simple unit tests for this?

If the strings contain multiple numbers at different places, it will use the first one for sorting right?

nkaenzig (commented):
Looks very nice! Left some minor comments.

configs/vision/dino_vit/offline/lits.yaml
```diff
@@ -0,0 +1,134 @@
...
+      train:
+        batch_size: &BATCH_SIZE ${oc.env:BATCH_SIZE, 64}
+        shuffle: true
+      val:
+        batch_size: *BATCH_SIZE
+      test:
+        batch_size: *BATCH_SIZE
+      predict:
+        batch_size: &PREDICT_BATCH_SIZE ${oc.env:PREDICT_BATCH_SIZE, 64}
+        # num_workers: 0
```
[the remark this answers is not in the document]
ioangatop: nice yes!

src/eva/vision/data/datasets/segmentation/lits.py
ioangatop: better yes

src/eva/core/utils/operations.py
ioangatop: it does not allow you to call the function with key-words; for example you can only do `numeric_sort(list)` and not `numeric_sort(item=list)`. This is useful for numeric / processes functions, like for example `add` `devide` etc when you can not have a "nice" argument name or its protected (`_item`), so the person will not do `numeric_sort(_item=list)`

ioangatop: yes add couple of unit tests!

it will be like this: `["a1b2", "a1b10", "a1b3"]` -> `["a1b2", "a1b3", "a1b10"]`

src/eva/vision/data/datasets/segmentation/lits.py
ioangatop: this is something that we have to sign, as in the other datasets the background is not one of the classes - its getting a bit weirded when the mask is not semantic labels (HxW) but a concat of binary one-hot (NUM_CLASSESxHxW), then the first channel represents the class 😅 . background is the absents of class so its not defined as a class if this makes sense, its always the 0 and datasets dont really refer to it

but we do want a consisted behaviour here, so probably I would say to include the background in all, but let me see how other are handling this

ioangatop: I added as background for now, I'll open an issue to discuss more about it

src/eva/core/utils/operations.py
nkaenzig: Would be good to add a test case like `["a1b2", "a1b10", "a1b3", "a2b1", "a2b4"]`

nkaenzig (approved):

src/eva/core/utils/operations.py
ioangatop: added this exactly one

PR #590, opened 2024-07-29
589 fix dice loss function for ignore index argument

### LINKED ISSUE #589: Fix DICE loss function for `ignore_index` argument

ioangatop:
Closes #589


roman807 (approved):
thanks!

PR #598, opened 2024-08-05
Use `core.models.wrappers` API to load encoder models

### LINKED ISSUE #595: Unify encoder API among classification & segmentation
These classes seem to be highly redundant right now (the former are being used for classification, the latter for segmentation) and should be unified:
- src/eva/vision/models/networks/from_timm.py v.s. src/eva/vision/models/networks/encoders/from_timm.py
- src/eva/core/models/networks/wrappers/from_function.py  src/eva/vision/models/networks/encoders/wrapper.py
 
We should unify the `wrappers` API from core with the `encoders` from vision.

nkaenzig:
Closes #595


src/eva/core/models/modules/head.py
```diff
@@ -25,7 +25,7 @@ class HeadModule(module.ModelModule):
 
     def __init__(
         self,
-        head: Dict[str, Any] | MODEL_TYPE,
+        head: MODEL_TYPE | Dict[str, Any],
```
ioangatop, line 28: we shouldn't change the order as we want to prioritise `Dict` so the model to initialized at `configure_model` step, so in every run to have different set of init params

nkaenzig: Reverted this change now, I wasn't aware that this might have a functional impactt.

roman807 (approved):
LGTM, thanks

ioangatop (approved):

Issue #267, opened 2024-03-15
Fix dependency groups 


ioangatop:
Remove them or duplicate the dependencies, according to https://github.com/pdm-project/pdm/issues/775#issuecomment-986212817

PR #605, opened 2024-08-09
Update .lock file to solve `exceptiongroup` dependency issue in CI

### LINKED ISSUE #604: Fix CI issue (No module named `exceptiongroup`)
Getting this error in CI when using python 3.10 (probably related to new pytest version that uses python 3.11 features):
```
ModuleNotFoundError: No module named 'exceptiongroup'
```

Context: https://github.com/pytest-dev/pytest/issues/12177#issue-2220516002
- For some reason nox / pdm doesn't install the `exceptiongroup` dependency when using python 3.10

nkaenzig:
Closes #604 

Note: I had to regenerate the lock file from scratch.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,6 +112,8 @@ typecheck = [
 ]
 test = [
     "pygments>=2.14.0",
+    "tomli>=2.0.1",
+    "exceptiongroup>=1.2.2",
     "pytest>=7.2.2",
     "pytest-cov>=4.1.0",
 ]
@@ -132,7 +134,6 @@ dev = [
     "yamllint>=1.29.0",
     "bandit>=1.7.6",
     "pyright>=1.1.295",
-    "pytest>=7.2.2",
     "nox>=2024.3.2",
     "pygments>=2.14.0",
     "pytest>=7.2.2",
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -71,17 +71,12 @@ vision = [
...
-    "scikit-image>=0.24.0",
-    "imagesize>=1.4.1",
-    "scipy>=1.14.0",
-    "monai>=1.3.2",
+    "kaiko-eva[vision]",
+    "kaiko-eva[lint]",
+    "kaiko-eva[typecheck]",
+    "kaiko-eva[test]",
+    "kaiko-eva[docs]",
+    "kaiko-eva[dev]",
```
ioangatop, line 79: this actually is not entirely correct, as it points to the pip version of the lib, and not the current local one, so the dependencies are not the same

issue: https://github.com/kaiko-ai/eva/issues/267
comment: https://github.com/pdm-project/pdm/issues/775#issuecomment-986212817
feature: https://github.com/pdm-project/pdm/issues/2701

nkaenzig: Good point. Reverted this now to manually copying the requirements.

ioangatop (approved):

PR #607, opened 2024-08-12
Add support for running `timm` models with default configs

### LINKED ISSUE #606: Add support for running timm models with default configs
Currently, with the default config users can only load models that `eva` exposes through its model registry. It would be handy to add the functionality there to support running any `timm` model, without the need of copying and manually editing the `.yaml` configs first.

nkaenzig:
Closes #606 

Example usage:

```
MODEL_NAME=universal/timm_model \
MODEL_EXTRA_KWARGS='{model_name: vit_small_patch16_224.dino, checkpoint_path: null}' \
eva predict_fit --config configs/vision/pathology/offline/segmentation/consep.yaml
```

### DIFF
--- a/configs/vision/pathology/offline/classification/bach.yaml
+++ b/configs/vision/pathology/offline/classification/bach.yaml
@@ -30,7 +30,7 @@ trainer:
             0: train
             1: val
           backbone:
-            class_path: eva.vision.models.wrappers.VisionBackbone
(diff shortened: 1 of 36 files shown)


roman807:
Great, thanks @nkaenzig! Could you add it also to the documentation? (e.g. maybe even new page)

roman807 (approved):

ioangatop (approved):

PR #617, opened 2024-08-23
Make dataset root fully configurable through env variable

### LINKED ISSUE #616: Dataset `root` is not fully configurable for some default configs
Currently in some configs the `root: ` init arg of the dataset is defined as follows:
`root: ${oc.env:DATA_ROOT, ./data}/bach`

This has two issues:
- Expects that all datasets are in the same `DATA_ROOT` folder when running evaluations on multiple benchmark datasets
- Forces the user to name the directory that contains the dataset files as in the config files

nkaenzig:
Closes #616

### DIFF
--- a/configs/vision/pathology/offline/classification/panda.yaml
+++ b/configs/vision/pathology/offline/classification/panda.yaml
@@ -94,7 +94,7 @@ data:
       predict:
         - class_path: eva.vision.datasets.PANDA
           init_args: &PREDICT_DATASET_ARGS
-            root: ${oc.env:DATA_ROOT, ./data/panda}
+            root: ${oc.env:DATA_ROOT, ./data/panda/prostate-cancer-grade-assessment}
             sampler:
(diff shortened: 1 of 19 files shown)


configs/vision/pathology/offline/classification/panda.yaml
```diff
@@ -94,7 +94,7 @@ data:
       predict:
         - class_path: eva.vision.datasets.PANDA
           init_args: &PREDICT_DATASET_ARGS
-            root: ${oc.env:DATA_ROOT, ./data/panda/prostate-cancer-grade-assessment}
+            root: ${oc.env:DATA_ROOT, ./data/panda}
```
roman807, line 97: if you download the dataset from kaggle it will be in `panda/prostate-cancer-grade-assessment`. So might be better to leave the full path as default

nkaenzig: Good point, updated now.

roman807 (approved):

PR #621, opened 2024-08-26
Set multiprocessing start method to `spawn` for `EmbeddingsDataset`

### LINKED ISSUE #620: DataLoader worker issue in `predict_fit` for segmentation datasets
Getting the following error, when running segmentation tasks like consep & monusac:

```
Traceback (most recent call last):
  File "/home/ray/anaconda3/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 1133, in _try_get_data
    data = self._data_queue.get(timeout=timeout)
  File "/home/ray/anaconda3/lib/python3.10/queue.py", line 179, in get
    raise Empty
_queue.Empty

RuntimeError: DataLoader worker (pid(s) 3236, 3237, 3238) exited unexpectedly
```

This only happens on linux machines with A100 GPUs, not locally (MacOS & M1 chip).
Also only happens with the `predict_fit` command, running `predict` & `fit` separately works.

nkaenzig:
Closes #620

On local machine (macOS) it was working because there the default start method is set to `spawn`, while on the linux machine where the error occured default used by `torch.DataLoader` was `fork` (If you don't specify `multiprocessing_context` expliclitly when initializing a `DataLoader` it will use the OS default).
Using `fork` can be unsafe and lead to issues as deadlocks (see [context](https://stackoverflow.com/a/66113051)).

### DIFF
--- a/src/eva/core/data/datasets/embeddings.py
+++ b/src/eva/core/data/datasets/embeddings.py
@@ -66,7 +66,7 @@ class EmbeddingsDataset(base.Dataset, Generic[TargetType]):
 
         self._data: pd.DataFrame
 
-        multiprocessing.set_start_method("spawn", force=True)
+        self._set_multiprocessing_start_method()
 
     def filename(self, index: int) -> str:
         """Returns the filename of the `index`'th data sample.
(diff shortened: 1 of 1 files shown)


src/eva/core/data/datasets/embeddings.py
```diff
@@ -65,6 +66,8 @@ def __init__(
 
         self._data: pd.DataFrame
 
+        multiprocessing.set_start_method("spawn", force=True)
+
```
ioangatop, line 70: maybe we can turn it into a method `_set_multiprocessing_method`, where in the function doc you write a bit what it going on? like

```
def __init__(self, ...)
  self._set_multiprocessing_method()
  
def _set_multiprocessing_method(self):
  """
  On CUDA devices we need to use the `spawn` method etc...
  """
```

ioangatop (approved):
Crazy difficult fix, thanks a lot Nic 💪

src/eva/core/data/datasets/embeddings.py
nkaenzig: Good point. Now put this within a private method `_set_multiprocessing_start_method`.

PR #626, opened 2024-09-03
Set available auto download to `False` for `camelyon16`

nkaenzig:
Auto download is not available for `camelyon16`, but currently `check_dataset_exists` prints a message recommending setting the `DOWNLOAD_DATA` env variable.

### DIFF
--- a/src/eva/vision/data/datasets/classification/camelyon16.py
+++ b/src/eva/vision/data/datasets/classification/camelyon16.py
@@ -161,7 +161,7 @@ class Camelyon16(wsi.MultiWsiDataset, base.ImageClassification):
 
     @override
     def prepare_data(self) -> None:
-        _validators.check_dataset_exists(self._root, True)
+        _validators.check_dataset_exists(self._root, False)
 
         expected_directories = ["training/normal", "training/tumor", "testing/images"]
         for resource in expected_directories:


src/eva/vision/data/datasets/classification/camelyon16.py
```diff
@@ -161,7 +161,7 @@ def annotations(self) -> Dict[str, str]:
 
     @override
     def prepare_data(self) -> None:
-        _validators.check_dataset_exists(self._root, True)
+        _validators.check_dataset_exists(self._root, False)
```
ioangatop, line 164: maybe we should change the message? now if the dataset does not exists, what is the message that we are getting?

nkaenzig: We're getting this message, because the `download_available` arg was set to `True`. 

```
https://github.com/kaiko-ai/eva/blob/b51c9ba00ce6ae44912f3db09492d3f69585b38d/src/eva/vision/data/datasets/_validators.py#L58
```

ioangatop: ah you are right! actually i wasnt aware about this addition 🙏

ioangatop (approved):

PR #635, opened 2024-09-11
Add init argument to freeze the encoder in `SemanticSegmentationModule`

### LINKED ISSUE #634: Freeze of encoder in `SemanticSegmentationModule` is not configurable
As for now, `SemanticSegmentationModule` will always freeze the encoder. However there are usecases where we want to train both the encoder & decoder, so this should be configurable.

nkaenzig:
Closes #634

### DIFF
--- a/src/eva/vision/models/modules/semantic_segmentation.py
+++ b/src/eva/vision/models/modules/semantic_segmentation.py
@@ -22,9 +22,10 @@ class SemanticSegmentationModule(module.ModelModule):
 
     def __init__(
         self,
-        decoder: decoders.Decoder,
+        decoder: decoders.Decoder | nn.Module,
         criterion: Callable[..., torch.Tensor],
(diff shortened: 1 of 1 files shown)


src/eva/vision/models/modules/semantic_segmentation.py
```diff
@@ -22,9 +22,10 @@ class SemanticSegmentationModule(module.ModelModule):
 
     def __init__(
         self,
-        decoder: decoders.Decoder,
+        decoder: decoders.Decoder | nn.Module,
         criterion: Callable[..., torch.Tensor],
         encoder: Dict[str, Any] | Callable[[torch.Tensor], List[torch.Tensor]] | None = None,
+        freeze_encoder: bool = False,
```
ioangatop, line 28: I believe that the encoder freezes when `lr_multiplier_encoder=0`. so we can escape an additional argument

nkaenzig: Oh right, missed that.

PR #636, opened 2024-09-11
Update `SemanticSegmentationModule`

nkaenzig:
- accept any `nn.Module` as decoder
- allow forward pass with a single argument to the forward pass

### DIFF
--- a/src/eva/vision/models/modules/semantic_segmentation.py
+++ b/src/eva/vision/models/modules/semantic_segmentation.py
@@ -22,7 +22,7 @@ class SemanticSegmentationModule(module.ModelModule):
 
     def __init__(
         self,
-        decoder: decoders.Decoder,
+        decoder: decoders.Decoder | nn.Module,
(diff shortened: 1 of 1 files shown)


roman807 (approved):

src/eva/vision/models/modules/semantic_segmentation.py
```diff
@@ -22,7 +22,7 @@ class SemanticSegmentationModule(module.ModelModule):
 
     def __init__(
         self,
-        decoder: decoders.Decoder,
+        decoder: decoders.Decoder | nn.Module,
```
ioangatop, line 25: idk if this is a good idea, as the decoder specifically wants the features to be 4D and the output is also very specific, which makes the Module a bit to loose and generic in my view

nkaenzig: Good point, but it allows people to use their own decoders without subclassing eva's `Decoder` class.

PR #645, opened 2024-09-24
Use `GridSampler` for `CoNSeP`

### LINKED ISSUE #644: Use `GridSampler` instead of `ForegroundGridSampler` for `CoNSeP`
The images in consep are already foreground patches, so we should extract all grid patches for all images.

nkaenzig:
Closes #644

### DIFF
--- a/src/eva/vision/data/datasets/segmentation/consep.py
+++ b/src/eva/vision/data/datasets/segmentation/consep.py
@@ -37,8 +37,8 @@ class CoNSeP(wsi.MultiWsiDataset, base.ImageSegmentation):
         root: str,
         sampler: samplers.Sampler | None = None,
         split: Literal["train", "val"] | None = None,
-        width: int = 224,
-        height: int = 224,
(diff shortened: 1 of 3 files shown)


ioangatop (approved):

src/eva/vision/data/datasets/segmentation/consep.py
```diff
@@ -37,8 +37,8 @@ def __init__(
         root: str,
         sampler: samplers.Sampler | None = None,
         split: Literal["train", "val"] | None = None,
-        width: int = 224,
-        height: int = 224,
+        width: int = 250,
+        height: int = 250,
```
ioangatop, line 41: lets do 256 or 224

nkaenzig: 250 is intentional here because the original image is 1000x1000, and we want to divide it into 4x4 patches.
Resisizing to 224 is done via a transformation afterwards.

PR #659, opened 2024-10-04
Add option to `LiTS` to adjust orientation

### LINKED ISSUE #658: CT scan orientation is wrong in `LiTS`
Currently the LiTS dataset classes return images like this:
image

So in some samples, the table is on the left side, and in others on the right side. Radiologist are used to have the table at the bottom, liver on the left, and heart on the right ("looking into the feet of the patient").

nkaenzig:
Closes #658

image

### DIFF
--- a/src/eva/vision/data/datasets/segmentation/lits.py
+++ b/src/eva/vision/data/datasets/segmentation/lits.py
@@ -29,6 +29,9 @@ class LiTS(base.ImageSegmentation):
     _test_ratio: float = 0.15
     """Index ranges per split."""
 
+    _fix_orientation: bool = True
+    """Whether to fix the orientation of the images to match the default for radiologists."""
+
     _sample_every_n_slices: int | None = None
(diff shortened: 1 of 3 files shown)


src/eva/vision/data/datasets/segmentation/lits.py
```diff
@@ -49,6 +51,7 @@ def __init__(
         root: str,
         split: Literal["train", "val", "test"] | None = None,
         transforms: Callable | None = None,
+        fix_orientation: bool = True,
```
ioangatop, line 54: maybe we should put it as a class state variable (like we do on the splits) because i think people wont want to change it

ioangatop (approved):

src/eva/vision/data/datasets/segmentation/lits.py
nkaenzig: done

PR #682, opened 2024-10-11
Add starplot leaderboard script to tools

roman807:
closes https://github.com/kaiko-ai/eva/issues/681


tools/generate_leaderboard_plots.py
```diff
@@ -0,0 +1,235 @@
...
+
+    # create colormap:
+    colors = [[0, "white"], [1, "#0000ff"]]
+    cmap = matplotlib.colors.LinearSegmentedColormap.from_list("", colors)
+
+    # prepare data frame:
+    df.index.names = [""]
+
+    # exclude BACH from the average
+    df_for_avg = df[[c for c in df.columns if c != "bach"]]
```
nkaenzig, line 134: should we specify a list above `_exclude_from_avg = ["bach"]` to specify datasets to exclude?

nkaenzig (approved):

tools/data/leaderboard.csv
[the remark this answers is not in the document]
roman807: The CSV that is being saved keeps the order of FMs as indicated in the hardcoded list. This makes it easier to control the color scheme for the starplot. The ordering according to avg performance is only needed for the visualization of the table

tools/generate_leaderboard_plots.py
roman807: done

PR #686, opened 2024-10-14
Add `CrossEntropyLoss` and `DiceCELoss` wrappers with support for class weights

### LINKED ISSUE #684: Add support for weighted loss functions
For imbalanced datasets adding class weights to the loss function can be important. However, as for now for instance the `weight` parameter in `CrossEntropyLoss` can't be use, because it expects a `torch.Tensor` which cannot be specified in the .yaml config / initialized by jsonargparse.

nkaenzig:
Closes #684


src/eva/vision/losses/dice.py
```diff
@@ -12,29 +14,94 @@ class DiceLoss(losses.DiceLoss):  # type: ignore
...
+        inputs, targets = _apply_ignore_index(inputs, targets, self.ignore_index)
+        targets = _to_one_hot(targets, num_classes=inputs.shape[1])
 
         if targets.ndim == 3:
             targets = one_hot(targets[:, None, ...], num_classes=inputs.shape[1])
 
         return super().forward(inputs, targets)
+
+
+class DiceCELoss(losses.dice.DiceCELoss):
```
roman807, line 54: no strong feelings but usually prefer spelling names out to make it more clear if looked at without context. Here could be `DiceCrossEntropyLoss`

roman807 (approved):

src/eva/vision/losses/dice.py
nkaenzig: Good point. In this instance I kept the original name, to make it more clear that those are just wrappers around commonly used loss functions from other libraries (monai in this case).

PR #690, opened 2024-10-18
Switch from `SGD` optimizer to `AdamW` for patch classification tasks 

nkaenzig:
- For some models we have observed SGD to not converge (fast enough), without further tuning the learning rate
- This PR standardises the batch size for patch classifiction tasks to 256


docs/leaderboards.md
```diff
@@ -38,26 +38,24 @@ We selected this approach to prioritize reliable, robust and fair FM-evaluation
 | **Dropout**                    | 0.0                       | 0.0                       | 0.0                       |
 | **Hidden activation function** | n/a                       | ReLU                      | n/a                       |
 | **Output activation function** | none                      | none                      | none                      |
-| **Number of steps**            | 12,500                    | 12,500 (2)                | 2,000                     |
```
roman807, line 41: Let's adjust the sentence in line 30?

Maybe something like 
```
*eva* uses a fixed protocol customized to each category of tasks. The setup has proven to be performant and robust independent of task and model size & architecture and generally prioritizes fairness and comparability over state-of-the-art performance.
```

(If you adopt this you can also remove the citation in line 61)

roman807:
Thanks, looks good! Do you want to already include the updated leaderboard plots in this PR or would this come in a later one?

docs/leaderboards.md
nkaenzig: Good point, updated now

roman807 (approved):

Issue #694, opened 2024-10-22
Add Virchow2 model to leaderboard


nkaenzig:
https://huggingface.co/paige-ai/Virchow2

PR #696, opened 2024-10-22
Add Virchow2 model to registry

nkaenzig:
Contributes to #694

### DIFF
--- a/src/eva/core/models/transforms/extract_cls_features.py
+++ b/src/eva/core/models/transforms/extract_cls_features.py
@@ -8,18 +8,18 @@ class ExtractCLSFeatures:
     """Extracts the CLS token from a ViT model output."""
 
     def __init__(
-        self, cls_index: int = 0, num_reg_tokens: int = 0, include_patch_tokens: bool = False
+        self, cls_index: int = 0, num_register_tokens: int = 0, include_patch_tokens: bool = False
     ) -> None:
         """Initializes the transformation.
 
         Args:
             cls_index: The index of the CLS token in the output tensor.
-            num_reg_tokens: The number of register tokens in the model output.
+            num_register_tokens: The number of register tokens in the model output.
             include_patch_tokens: Whether to concat the mean aggregated patch tokens with
                 the cls token.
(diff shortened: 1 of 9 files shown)


src/eva/core/models/transforms/extract_cls_features.py
```diff
@@ -7,13 +7,20 @@
 class ExtractCLSFeatures:
     """Extracts the CLS token from a ViT model output."""
 
-    def __init__(self, cls_index: int = 0) -> None:
+    def __init__(
+        self, cls_index: int = 0, num_reg_tokens: int = 0, include_patch_tokens: bool = False
```
roman807, line 11: lets rename `num_reg_tokens` -> `num_register_tokens`? (reg could also mean something else)

src/eva/vision/models/networks/backbones/pathology/histai.py
```diff
@@ -42,5 +42,5 @@ def histai_hibou_l(out_indices: int | Tuple[int, ...] | None = None) -> nn.Modul
         model_name="histai/hibou-L",
         out_indices=out_indices,
         model_kwargs={"trust_remote_code": True},
-        transform_args={"ignore_remaining_dims": True} if out_indices is not None else None,
+        transform_args={"num_reg_tokens": 4} if out_indices is not None else None,
```
roman807, line 45: maybe make this a variable, or explain in a comment where "4" is coming from?

src/eva/core/models/transforms/extract_cls_features.py
nkaenzig: done

src/eva/vision/models/networks/backbones/pathology/histai.py
nkaenzig: This needs to be a constant, is a hyperparameter of the ViT (their models use 4 register tokens).

nkaenzig: Added a comment now to the docstring.

PR #701, opened 2024-11-07
Bump version to `0.1.2`

nkaenzig:


pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = "pdm.backend"
 
 [project]
 name = "kaiko-eva"
-version = "0.1.0"
+version = "0.1.2"
```
roman807, line 9: should be "0.1.1" first?

roman807 (approved):

pyproject.toml
nkaenzig: No, the current version on `pypi` is `0.1.1` already, I'm not sure why in the `.toml` it was still at `0.1.0`

PR #711, opened 2024-11-21
Update leaderboard with new segmentation results & adding new FMs

nkaenzig:
- Adding Virchow2 & Phikon-v2
- Updating segmentation metrics (consep, monusac) after rerunning using a new conv decoder that also receives the original input image in addition to the FM feature maps. This reduces the previous unfair advantage that FMs with smaller patch sizes possessed for segmentation.


docs/images/leaderboard.svg
roman807, line 1: is the ranking computed consistently with the internal W&B leaderboard? here i see prov-gigapath ahead of UNI in W&B we have it the other way around

nkaenzig: The numbers of the metrics are exactly the same. seems that in the internal script we apply a round to 3 digits before sorting, and UNI & Prov-Gigapath differ in the 4th digit. In this public leaderboard we don't round the computed average, so this one is more reliable.

roman807 (approved):

PR #713, opened 2024-11-22
Add support for dataloader samplers

### LINKED ISSUE #712: Add support for label efficiency evals
One of the key advantages of FMs is that it has the potential to reduce the required number of annotated samples significantly. We should add some features to measure the label efficiency of FMs, for instance by subsampling the number of training samples.

nkaenzig:
Closes #712

- Adds a `samplers` argument to `eva.DataModule`, so we can enable custom samplers in the dataloaders.
  - This enables for instance label efficiency experiments, as we can reduce the number of training samples for the existing downstream tasks.
- Adds a `BalancedSampler` which supports balanced class sample data loading for classification tasks.  

## How to use

Just add this to the `init_args`  of the `DataModule` in your yaml config:

```
    samplers:
      train:
        class_path: eva.core.data.samplers.classification.BalancedSampler
        init_args:
          num_samples: 10
```

(For online mode, specify the sampler in `samplers.train:`, while for offline mode in `samplers.predict:`).

Make sure that `shuffle: false` for the dataloader config of the corresponding split.


tests/eva/core/models/modules/test_inference.py
```diff
@@ -19,7 +19,7 @@
     "dataset_fixture",
     [
         "classification_dataset",
-        "classification_dataset_with_metadata",
+        # "classification_dataset_with_metadata",
```
roman807, line 22: why is this outcommented?

roman807 (approved):

tests/eva/core/models/modules/test_inference.py
nkaenzig: good catch, now uncommented it

PR #719, opened 2024-11-28
Replace `GeneralizedDiceScore` by `DiceScore` & fix class-wise metrics

### LINKED ISSUE #718: Fix classwise dice metrics
For calculating the classwise dice metrics, we currently use `torchmetrics.segmentation.GeneralizedDiceScore`, which in this context should be equivalent to standard dice score. (Standard `Dice` so far did not support classwise calculation, because the `average='none'` option was broken (see [issue](https://github.com/Lightning-AI/torchmetrics/issues/2262)).

However, one issue with the current implementation of `GeneralizedDiceScore` in torchmetrics (`<1.6.0`) is that it assigns scores of `0.0` to samples that don't contain a particular class, which leads to very low classwise dice scores, even if the predictions are of good quality.

Seems that this issue is fixed in the new `torchmetrics.segmentation.DiceScore` which has been released in torchmetrics 1.6.0, so we should update to that version and move away from `GeneralizedDiceScore`.

nkaenzig:
Closes #718

Used `DiceMetric` from MONAI for class-wise scores, because in torchmetrics there is an open issue for this:
https://github.com/Lightning-AI/torchmetrics/issues/2851

### DIFF
--- a/configs/vision/pathology/online/segmentation/consep.yaml
+++ b/configs/vision/pathology/online/segmentation/consep.yaml
@@ -21,7 +21,7 @@ trainer:
           filename: best
           save_last: true
           save_top_k: 1
-          monitor: &MONITOR_METRIC ${oc.env:MONITOR_METRIC, val/GeneralizedDiceScore}
+          monitor: &MONITOR_METRIC ${oc.env:MONITOR_METRIC, 'val/DiceScore (micro)'}
           mode: &MONITOR_METRIC_MODE ${oc.env:MONITOR_METRIC_MODE, max}
       - class_path: lightning.pytorch.callbacks.EarlyStopping
         init_args:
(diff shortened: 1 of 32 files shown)


configs/vision/pathology/online/segmentation/consep.yaml
```diff
@@ -79,11 +79,17 @@ model:
...
                 num_classes: *NUM_CLASSES
-                weight_type: linear
-                per_class: true
+                reduction: none
+            labels:
+              - background
+              - other
+              - inflammatory
+              - epithelial
+              - spindle-shaped
```
roman807, line 92: are these all labels for consep? in the doc we mention 7 nuclei types: https://kaiko-ai.github.io/eva/main/datasets/consep/

configs/vision/pathology/online/segmentation/monusac.yaml
```diff
@@ -81,12 +81,19 @@ model:
...
+                reduction: none
                 ignore_index: *IGNORE_INDEX
-                per_class: true
+            labels:
+              - background
+              - epithelial
+              - lymphocyte
+              - neutrophil
+              - macrophage
+              - ambiguous
```
roman807, line 96: also here, num classes is inconsistent with documentation. https://kaiko-ai.github.io/eva/main/datasets/monusac/ If doc is wrong can you fix it or create a separate issue? (maybe we should also mention the classes in the doc)

roman807 (approved):
LGTM, maybe @ioangatop should have a look as well here

configs/vision/pathology/online/segmentation/consep.yaml
nkaenzig: good catch, updated the docs now

configs/vision/pathology/online/segmentation/monusac.yaml
nkaenzig: updated the docs

PR #724, opened 2024-12-05
Override checkpointing hooks to exclude backbone while saving checkpoints

### LINKED ISSUE #723: `ModelCheckpoint` callback saves the encoder weights even though they are frozen
During evals, we currently use `lightning.pytorch.callbacks.ModelCheckpoint` to save the best model checkpoints during `fit`. While we only fit the decoder / head on the downstream tasks, the encoder usually remains frozen, so it's not necessary to include it into the saved checkpoints.

Furthermore, for big encoders (e.g. ViT-G), this becomes a major bottleneck in terms of runtime, taking up to 30 seconds to save the checkpoint, while the GPU remains idle:
Image

nkaenzig:
Closes #723

### DIFF
--- a/src/eva/core/models/modules/head.py
+++ b/src/eva/core/models/modules/head.py
@@ -12,7 +12,7 @@ from typing_extensions import override
 from eva.core.metrics import structs as metrics_lib
 from eva.core.models.modules import module
 from eva.core.models.modules.typings import INPUT_BATCH, MODEL_TYPE
-from eva.core.models.modules.utils import batch_postprocess, grad
+from eva.core.models.modules.utils import batch_postprocess, grad, submodule_state_dict
 from eva.core.utils import parser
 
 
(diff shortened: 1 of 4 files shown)


src/eva/core/models/modules/head.py
```diff
@@ -32,6 +32,7 @@ def __init__(
         lr_scheduler: LRSchedulerCallable = lr_scheduler.ConstantLR,
         metrics: metrics_lib.MetricsSchema | None = None,
         postprocess: batch_postprocess.BatchPostProcess | None = None,
+        save_head_only: bool = True,
```
ioangatop, line 35: maybe we should have it as class attribute?

```
class Module:
   save_head_only: bool = True
   """bla bla bla"""
```

ioangatop (approved):

PR #730, opened 2024-12-09
CI documentation fixing

ioangatop:


pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = "pdm.backend"
 
 [project]
 name = "kaiko-eva"
-version = "0.1.5"
+version = "0.1.7"
```
nkaenzig, line 9: this should be `0.1.6`

nkaenzig: (for some reason the CI releases always a version +1 with respect to what is specified here, so if you set it to `0.1.7` it would release `0.1.8`, and skipping `0.1.7`. See also release history, where some versions are missing due to this: https://pypi.org/project/kaiko-eva/#history)

ioangatop: pypi is `0.1.6` 🤔

ioangatop: ok done!

nkaenzig (approved):

PR #734, opened 2024-12-10
Change `GeneralizedDiceScore` to `MonaiDiceScore` metric in leaderboard

nkaenzig:
In https://github.com/kaiko-ai/eva/pull/719 we added support for the monai dice metric to eva, which seems to be more reliable than the current `torchmetrics` implementations.


nkaenzig:
done in https://github.com/kaiko-ai/eva/pull/772

PR #740, opened 2024-12-27
Add PubMedQA dataset

kurbanrita:
Add PubMedQA dataset

PubMedQA is a biomedical question-answering dataset collected from PubMed abstracts. The task of PubMedQA is to answer research questions with yes/no/maybe. The subset that is used for validation has 1k expert-annotated QA instances.

- [x] Add LanguageDataset
- [x] Add TextClassification dataset
- [x] Add PubMedQA dataset that is loaded from HuggingFace (incl. add HuggingFace `datasets` as a new language dependency)
- [x] Add tests

Remaining questions:
- Vision validators could probably be reused for language. Maybe refactor some common functions into `core`?
- The PR won't be merged to main, the development will continue on this branch.


src/eva/language/data/datasets/classification/base.py
```diff
@@ -0,0 +1,66 @@
...
+    """Text classification abstract dataset."""
+
+    def __init__(self) -> None:
+        """Initializes the text classification dataset."""
+        super().__init__()
+
+    @property
+    def classes(self) -> List[str] | None:
+        """Returns list of class names."""
+        return None
```
ioangatop, line 22: we can remove the `return None` here all together (same to the other ones) - modern python code neglect them

src/eva/language/data/datasets/language.py
```diff
@@ -0,0 +1,15 @@
...
+from eva.core.data.datasets import base
+
+DataSample = TypeVar("DataSample")
+"""The data sample type."""
+
+
+class LanguageDataset(base.MapDataset, abc.ABC, Generic[DataSample]):
+    """Base dataset class for text tasks."""
+
+    pass
```
ioangatop, line 15: we can remove the `pass` here (modern python code neglect them)

src/eva/language/data/datasets/classification/pubmedqa.py
```diff
@@ -0,0 +1,126 @@
...
+class PubMedQA(base.TextClassification):
+    """Dataset class for PubMedQA question answering task."""
+
+    _license: str = "MIT License (https://github.com/pubmedqa/pubmedqa/blob/master/LICENSE)"
+    """Dataset license."""
+
+    def __init__(
+        self,
+        root: str | None = None,
+        split: str | None = "train+test+validation",
```
ioangatop, line 22: Lets use Literal type here; `split: Literal["train", "val", "test"] | None = None` and when `None` it will use all of them

ioangatop (commented):
Great work @kurbanrita 🎉  here are some initial comments

ioangatop (commented):
Some more comments 🤗

src/eva/language/data/datasets/classification/pubmedqa.py
```diff
@@ -0,0 +1,130 @@
...
+            split: Valid splits among ["train", "validation", "test"].
+                If None, it will use "train+test+validation".
+            download: Whether to download the dataset if not found locally. Default is False.
+        """
+        super().__init__()
+        self._root = root
+        if split is None:
+            self._split = "train+test+validation"
+        else:
+            self._split = "+".join(split)
```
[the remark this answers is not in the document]
kurbanrita: ok, HuggingFace expects arguments such as "train+test+validation", "train+validation" or "train", so maybe I could simply specify this format in the doc string because otherwise the most intuitive thing will be to pass ['train', 'test'], which is not what `load_dataset` expects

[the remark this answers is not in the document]
kurbanrita: since I don't foresee us using any other splits except the default, it shouldn't be an issue

src/eva/language/data/datasets/classification/pubmedqa.py
```diff
@@ -0,0 +1,138 @@
...
+class PubMedQA(base.TextClassification):
+    """Dataset class for PubMedQA question answering task."""
+
+    _license: str = "MIT License (https://github.com/pubmedqa/pubmedqa/blob/master/LICENSE)"
+    """Dataset license."""
+
+    def __init__(
+        self,
+        root: Optional[str] = None,
+        split: Literal["train", "validation", "test"] | None = None,
```
nkaenzig, line 23: nit: In other dataset classes we use `"val"` (not `"validation"`) to refer to the validation split.

src/eva/language/data/datasets/classification/pubmedqa.py
```diff
@@ -0,0 +1,138 @@
...
+                If None, it will use "train+test+validation".
+            download: Whether to download the dataset if not found locally. Default is False.
+        """
+        super().__init__()
+
+        self._root = root
+        self._split = split
+        self._download = download
+
+    def _load_dataset(self, dataset_cache_path: Optional[str]) -> Dataset:
```
nkaenzig, line 40: nit: Usually we use `str | None` instead of `Optional[str]`

nkaenzig (approved):
Very nice work @kurbanrita! Left a couple of nitpicks, after those I think this should be ready to go.

ioangatop (approved):
Great start @kurbanrita 🎉  Lets address @nkaenzig comments and merge :D

src/eva/language/data/datasets/classification/pubmedqa.py
kurbanrita: this is what HuggingFace uses, I could map it to "validation" internally but for evaluation we will always use the default because this is already a human-annotated test set, so maybe just remove the parameter altogether?

nkaenzig: I would map it to "validation" when doing the call to huggingface's load dataset function and use `val` here to keep the "interface" among eva's dataset classes consistent.

Issue #739, opened 2024-12-24
Integrate NLP Modality into EVA


kurbanrita:
This is an epic ticket where I will document the integration of the NLP modality to eva. 

High-Level Tasks:

- [ ] Literature review of relevant datasets that can be used as eva tasks.
- [ ] Adding a custom text dataset to a new language folder.
- [ ] Define relevant metrics for text classification, free-form text generation, etc. 
- [ ] End-to-end pipeline set up, including relevant modificaitons to the trainer, configurations, etc.

kurbanrita:
The dataset literature review will be updated here: https://kaiko-ai.atlassian.net/l/cp/Xg1JNjZJ.

I have preliminarily chosen to use [PubMedQA](https://pubmedqa.github.io) as the first text task for eva. First, its classification task with yes/no/maybe questions maps naturally to eva’s classification task, ensuring seamless integration. PubMedQA provides over 1,000 high-quality, human-annotated questions, offering reliable and accurate benchmarks. Being fully open-source, it allows easy access and use without licensing issues. The QA format aligns perfectly with our internal focus on question–answer tasks, making it highly relevant to our needs. Additionally, PubMedQA is not too easy, providing a meaningful challenge that effectively differentiates between simple models and advanced LLMs. Finally, as a widely used and trusted dataset in the medical NLP community, PubMedQA ensures that our evaluations are based on a credible and respected benchmark. These factors make PubMedQA an ideal initial choice for expanding eva’s capabilities to include text-based evaluations.

PR #758, opened 2025-02-05
Add end-to-end example for `language` modality using HuggingFace

kurbanrita:
Add functionality to run HuggingFace models on text data.

- [x] Add validate function to the interface that runs validation without fitting
- [x] Add TextModule for LLMs
- [x] Add HuggingFace wrapper
- [x] Add configuration for a simple validation task
- [x] Add tests


src/eva/language/data/datasets/classification/pubmedqa.py
```diff
@@ -118,15 +118,18 @@ def load_target(self, index: int) -> torch.Tensor:
...
-            "reasoning_free": sample["reasoning_free_pred"],
+            key: str(value) if value is not None else ""
+            for key, value in {
+                "year": sample.get("YEAR"),
+                "labels": sample.get("LABELS"),
+                "meshes": sample.get("MESHES"),
+                "long_answer": sample.get("LONG_ANSWER"),
+                "reasoning_required": sample.get("reasoning_required_pred"),
+                "reasoning_free": sample.get("reasoning_free_pred"),
+            }.items()
```
ioangatop, line 132: We can keep it simple
```
return {
    "year": sample.get("YEAR", ""),
    "labels": sample.get("LABELS", ""),
    ...
}
```

or more complicated
```
return {
    key.lower().removesuffix("_pred"): sample.get(key, "")
    for key in ("YEAR", "LABELS", ...)
}
```

IMO simple is enough and more readable :)

src/eva/core/trainers/functional.py
```diff
@@ -131,3 +131,32 @@ def infer_model(
         datamodule=datamodule,
         return_predictions=return_predictions,
     )
+
+
+def run_validation_only(
```
ioangatop, line 136: just `run_validation` its a good name

ioangatop (commented):
Great work! this is some first review mostly about styling

src/eva/language/data/datasets/classification/pubmedqa.py
kurbanrita: oh, sure, makes sense

kurbanrita: I tested it and if a key exists in the sample but the value is None, then `sample.get(key, "")` will still return None, which is what I try to avoid. Maybe I'll use `"year": sample.get("YEAR") or "",`? what do you think?

src/eva/core/trainers/functional.py
```diff
@@ -131,3 +131,32 @@ def infer_model(
         datamodule=datamodule,
         return_predictions=return_predictions,
     )
+
+
+def run_validation(
```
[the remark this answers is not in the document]
nkaenzig: I think it would cleaner to integrate this into `run_evaluation_session`, one option could be to add a `stages: List[Literal["fit", "validate", "test"]]`, and then within `run_evaluation_session` we have logic that controls which stages are executed. Any thoughts here @ioangatop ?

src/eva/core/trainers/trainer.py
```diff
@@ -113,3 +113,26 @@ def run_evaluation_session(
...
+
+        Args:
+            model: The model to validate (cloned, not modified in-place).
+            datamodule: The datamodule for validation.
+            verbose: Whether to print validation metrics to stdout.
+
+        Returns:
+            Validation metrics as returned by trainer.validate.
+        """
+        functional.run_validation(
```
nkaenzig, line 133: If we do the change mentioned above to `run_evaluation_session`, we could remove `validate_only` here?

src/eva/language/models/networks/module.py
```diff
@@ -0,0 +1,86 @@
...
+
+        Returns:
+            Dictionary with predictions, ground truth, and evaluation metrics.
+        """
+        data, targets, metadata = INPUT_BATCH(*batch)
+        messages = [self.prompt + "\n" + d + "\nAnswer: " for d in data]
+        predictions = self(messages)
+        logger.info(f"Predictions: {predictions}")
+        logger.info(f"Targets: {targets}")
+        # TODO: Add support for evaluation metrics
```
nkaenzig, line 85: Can we open an issue for this?

src/eva/language/models/networks/module.py
```diff
@@ -0,0 +1,86 @@
...
+        Args:
+            batch: A batch containing 'QUESTION', 'CONTEXTS', 'final_decision', etc.
+
+        Returns:
+            Dictionary with predictions, ground truth, and evaluation metrics.
+        """
+        data, targets, metadata = INPUT_BATCH(*batch)
+        messages = [self.prompt + "\n" + d + "\nAnswer: " for d in data]
+        predictions = self(messages)
+        logger.info(f"Predictions: {predictions}")
```
nkaenzig, line 83: We might want to remove the `logger.info` calls here or change to `logger.debug` as this could quickly bloat the logs?

src/eva/language/models/wrappers/huggingface.py
```diff
@@ -0,0 +1,63 @@
...
+            generate_kwargs: Additional generation parameters (temperature).
+
+        Returns:
+            The generated text as a string.
+        """
+        outputs = self._pipeline(prompts, return_full_text=False, **generate_kwargs)
+        if outputs is None:
+            raise ValueError("Outputs from the model are None.")
+        return [
+            output[0]["generated_text"] if isinstance(output, list) else output
```
nkaenzig, line 61: Does `output[0]["generated_text"]` hold for all models to get the generated text output?

nkaenzig (commented):
Looks very good! I left a couple of comments.

src/eva/language/models/networks/module.py
kurbanrita: agree!

src/eva/language/models/networks/module.py
kurbanrita: I am working on a separate language branch, the steps are outlined here: https://github.com/kaiko-ai/eva/issues/739

src/eva/language/models/wrappers/huggingface.py
kurbanrita: We currently support only `text_generation` task as that's how modern LLMs operate. It always returns output in this format! I will update it if there is a need

kurbanrita: I saw that there was a text_classification task too in the literal, removed it for now!

src/eva/core/trainers/functional.py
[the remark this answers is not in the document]
ioangatop: i agree that's good let's do it

src/eva/language/models/wrappers/huggingface.py
nkaenzig: I see, I thought `output[0]["generated_text"]` might be something coming from the forward pass implementation of a specific model. But if this is used in HuggingFace accross all text generation pipelines it should be fine.

kurbanrita: Implemented, could you please review @nkaenzig?

src/eva/core/trainers/functional.py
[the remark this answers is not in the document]
kurbanrita: should we add test to validate too to make it analogous to fit but without the fitting part?
```
def validate():
  trainer.run_evaluation_session(model=model, datamodule=data, stages=["validate", "test"])
  ```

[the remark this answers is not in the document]
Comment (kurbanrita) [reply src/eva/core/trainers/functional.py]: Implemented, could you please review @nkaenzig?

File: src/eva/core/interface/interface.py [hunk 8]
```diff
@@ -34,7 +34,14 @@ def fit(
             model: The model module to use but not modify.
             data: The data module.
         """
-        trainer.run_evaluation_session(model=model, datamodule=data)
+        eva_trainer.run_evaluation_session(
+            base_trainer=trainer,
+            base_model=model,
+            datamodule=data,
+            stages=["fit", "validate", "test"],
+            n_runs=1,
```
nkaenzig, line 42: Why hardcoding number of runs to 1 here?

src/eva/core/interface/interface.py
```diff
@@ -77,3 +84,29 @@ def predict_fit(
...
+            trainer: The base trainer to use but not modify.
+            model: The model module to use but not modify.
+            data: The data module containing validation data.
+        """
+        eva_trainer.run_evaluation_session(
+            base_trainer=trainer,
+            base_model=model,
+            datamodule=data,
+            stages=["validate"],
+            n_runs=1,
```
nkaenzig, line 110: Same comment as above. (I think also for validate/ LLM evals multiple runs could make sense, especially when the model is configured in a way such that it produces different responses when querying it with the same prompt multiple times)

src/eva/core/interface/interface.py
```diff
@@ -34,7 +34,14 @@ def fit(
             model: The model module to use but not modify.
             data: The data module.
         """
-        trainer.run_evaluation_session(model=model, datamodule=data)
```
nkaenzig, line 37: Why changing here the function to the functional api? This way we lose run related settings that are stored within the trainer (e.g. `trainer.n_runs`)
Related to https://github.com/kaiko-ai/eva/pull/758#discussion_r2073670897:

nkaenzig: After giving this another thought, I think we could remove `run_evaluation_session` from the `Trainer` class completely, and always call the function from the functional API. We just need to make sure that we pass `trainer.n_runs` as arg.

PR #772, opened 2025-03-17
Increase patience to 20% and increase `n_runs` to 20 for slide level classification

nkaenzig:
- With 5% patience, in rare occurrences some eval runs wouldn't fully converge, therefore we increase patience to 20%.
- Standard deviation among runs for slide level classification tasks can be high, therefore we increase the number of runs from 5 to 20 to get a more accurate estimate of the average score (2x tighter confidence interval).
- Removed LR scheduler, as it was wrongly configured previously and doesn't seem to impact the evaluations much, especially when using an adaptive optimizers as adam


roman807 (approved):
looks good, thanks!

PR #773, opened 2025-03-21
Add `PatchCamelyon` 10-shot task config

karasikov:
- Added few-shot PatchCamelyon task (10 patches per class, 20 in total), 50 runs sampling different sequences every time
- Updated `BalancedSampler` logic, such that it samples different indices in each evaluation run


configs/vision/pathology/offline/classification/bach.yaml
```diff
@@ -17,7 +17,7 @@ trainer:
       - class_path: lightning.pytorch.callbacks.ModelCheckpoint
         init_args:
           filename: best
-          save_last: ${oc.env:SAVE_LAST, false}
+          save_last: true
```
nkaenzig, line 20: Why setting save_last to true and removing the omegaconf template placeholder `${oc.env:SAVE_LAST`? There are a couple of occurences of this in the PR.

configs/vision/pathology/offline/classification/camelyon16_small_mpp0.5.yaml
```diff
@@ -0,0 +1,141 @@
...
+        - class_path: eva.vision.datasets.Camelyon16
+          init_args: &PREDICT_DATASET_ARGS
+            root: ${oc.env:DATA_ROOT, ./data/camelyon16}
+            sampler:
+              class_path: eva.vision.data.wsi.patching.samplers.ForegroundGridSampler
+              init_args:
+                max_samples: *N_PATCHES
+            width: 448
+            height: 448
+            target_mpp: 0.25
```
nkaenzig, line 111: mpp here is set to 0.25?

src/eva/core/data/samplers/classification/balanced.py
```diff
@@ -78,7 +68,10 @@ def _make_indices(self):
         for idx in tqdm(
             range(len(self.data_source)), desc="Fetching class indices for balanced sampler"
         ):
-            _, target, _ = DataSample(*self.data_source[idx])
+            try:
+                target = self.data_source._load_target(idx)
```
nkaenzig, line 72: why is this needed?

src/eva/core/data/samplers/classification/balanced.py
```diff
@@ -94,3 +87,14 @@ def _make_indices(self):
                         f"Class {class_idx} has only {len(indices)} samples, "
                         f"which is less than the required {self._num_samples} samples."
                     )
+
+        # sample a new sequence
+        self._indices = []
```
nkaenzig, line 92: I'm not sure how the resampling of the indices is triggered in different runs of an evaluation?

karasikov: yes, you can compare with and without this patch, and print generated sequences each time

src/eva/core/data/samplers/classification/balanced.py
karasikov: Here we only need the target, and some datasets support this method that loads only the target. This avoids the overhead of loading the image, which is thrown away in https://github.com/kaiko-ai/eva/pull/773/files#diff-7f682f1a6ace5b924c50efd3bd93bf40ad3bf88a2ab77c1e8bb44fed6890801bR74
When this is supported, it is much-much faster

configs/vision/pathology/offline/classification/bach.yaml
karasikov: fine either way for me

configs/vision/pathology/offline/classification/camelyon16_small_mpp0.5.yaml
nkaenzig: as discussed, we'll exclude the cam16 0.5 config from this PR

nkaenzig (approved):

PR #785, opened 2025-03-31
Add `BTCV` dataset / support for 3D radiology datasets

### LINKED ISSUE #784: Add support for 3D radiology datasets
Right now, the radiology dataset that we support (e.g. `LiTS`, `TotalSegmentator`) are implemented in 2D fashion, meaning that we extract single slices as data samples.

This has multiple issues:
- Can't be used to evaluate 3D models
- For some datasets like `KiTS23`, loading single slices is very inefficient, as the slices are not always stored as consecutive blocks in the nifti files. So often loading a single slice requires basically loading the full nifti file content from disk.

nkaenzig:
Closes #784

- Refactoring of `VisionDataset` base class. Now all vision datasets use this as a base class (previously we had additional `ImageClassification` and `ImageSegmentation` base classes which were highly redundant), and the data types of the images & targets can be configured through TypeVars. Example:
  - **image classification / regression**
  ```python
  class ExampleDataset(vision.VisionDataset[tv_tensors.Image, torch.Tensor]): 
  ...
  ```
  
  - **semantic segmentation (2D)**
  ```python
  class ExampleDataset(vision.VisionDataset[tv_tensors.Image, tv_tensors.Mask]): 
  ...
  ```
  
  - **semantic segmentation (3D)**
  ```python
  class ExampleDataset(vision.VisionDataset[tv_tensors.Volume, tv_tensors.Mask]): 
  ...
  ```

- Added `BTCV` as a first 3D dataset example.


ioangatop (approved):
Simplily lovely @nkaenzig 🎉

PR #840, opened 2025-06-30
Update `language` wrappers to use `model.forward()` with the updated types

kurbanrita:
- [x] This PR updates the `language` module wrappers to use `model.forward` logic instead of `model.generate` as the typing has become more flexible after merging #839.
- [x] All tests are updated too. 
- [x] I have noticed that the HuggingFace wrapper had `transformers` versioning issues, added a comment to the tutorial.

### DIFF
--- a/src/eva/language/models/wrappers/huggingface.py
+++ b/src/eva/language/models/wrappers/huggingface.py
@@ -60,7 +60,10 @@ class HuggingFaceTextModel(base.BaseModel[List[str], List[str]]):
         outputs = self._pipeline(prompts, return_full_text=False, **self._generation_kwargs)
         if outputs is None:
             raise ValueError("Outputs from the model are None.")
-        return [  # type: ignore
-            output[0]["generated_text"] if isinstance(output, list) else output["generated_text"]
-            for output in outputs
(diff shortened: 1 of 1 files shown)


tests/eva/language/models/wrappers/test_huggingface.py
```diff
@@ -46,10 +46,14 @@ def test_real_small_hf_model_generation(
...
-        model = HuggingFaceTextModel(model_name_or_path=model_name_or_path, task="text-generation")
+        model = HuggingFaceTextModel(
+            model_name_or_path=model_name_or_path,
+            task="text-generation",
+            generation_kwargs=generate_kwargs,
+        )
 
-        output1 = model.generate([prompt], **generate_kwargs)[0]
-        output2 = model.generate([prompt], **generate_kwargs)[0]
+        output1 = model.forward([prompt])[0]
```
nkaenzig, line 55: Instead of `model.forward([prompt])[0]` we can now write `model([prompt])[0]` (which should call the forward method).

nkaenzig (approved):
LGTM, left only a small comment.

PR #848, opened 2025-07-09
Remove `TotalSegmentator2D`

nkaenzig:
This is a leftover from the old 2D radiology pipeline which uses a dataset class that returns individual slices.
Our new 2D configs (see https://github.com/kaiko-ai/eva/pull/845) use datasets that load the full CT scan, and then the slice sampling happens in the data transforms.


ioangatop (approved):
finally :D

PR #853, opened 2025-07-16
Change datasets requirement to `datasets<4.0.0,>=2.19.0`

nkaenzig:
Previous requirement was too tight as `language` pipeline also seems to work with older versions.


MaxFeucht (approved):
LGTM

PR #864, opened 2025-07-30
Enable `prepare_data` calling across GPUs in DataModule

MaxFeucht:


src/eva/core/data/datamodules/datamodule.py
```diff
@@ -39,6 +39,7 @@ def __init__(
         self.datasets = datasets or self.default_datasets
         self.dataloaders = dataloaders or self.default_dataloaders
         self.samplers = samplers or self.default_samplers
+        self.prepare_data_per_node = True
```
nkaenzig, line 42: We should expose this in the constructor such that we can configure it through the yaml configs. For the default value probably better if we keep lightning's default setting.

MaxFeucht:
Closing this PR since the desired effect of this PR could be achieved in a subclass-specific change

PR #866, opened 2025-08-11
Add `seed` argument to RandomSampler

MaxFeucht:


src/eva/core/data/samplers/random.py
```diff
@@ -10,24 +11,32 @@
...
+    """Samples elements randomly from a MapDataset."""
 
     data_source: datasets.MapDataset  # type: ignore
 
     def __init__(
-        self, replacement: bool = False, num_samples: Optional[int] = None, generator=None
+        self,
+        replacement: bool = False,
+        num_samples: Optional[int] = None,
+        generator: Optional[torch.Generator] = None,
```
nkaenzig, line 22: It might suffice if we just expose `seed` as an init arg? (I can't think of use cases where we would pass both).

And then something like,

```python
self._num_samples = num_samples
self._generator = None

if seed is not None:
    self._generator = torch.Generator()
    self._generator.manual_seed(seed)
```

PR #873, opened 2025-09-01
Add `offline` mode to language and multimodal

nkaenzig:


src/eva/language/callbacks/writers/prediction.py
```diff
@@ -0,0 +1,176 @@
...
+
+        prediction_batch = self._get_predictions(batch)
+
+        target_batch, prediction_batch = self._apply_postprocess(
+            pl_module, target_batch, prediction_batch
+        )
+
+        for i in range(len(batch_indices)):
+            entry: ManifestEntry = {
+                "text": message_utils.serialize(text_batch[i]),
```
MaxFeucht, line 108: Why do we need to serialize here?

src/eva/language/data/datasets/prediction.py
```diff
@@ -0,0 +1,148 @@
...
+    def load_text(self, index: int) -> MessageSeries | None:
+        """Returns the text for the given index."""
+        if self.text_column is None:
+            return None
+
+        text = self._data.iloc[index][self.text_column]
+
+        try:
+            return message_utils.deserialize(self._data.iloc[index][self.text_column])
+        except Exception:
```
MaxFeucht, line 117: This is an unspecific catch, why do we expect an Exception here and can we document this?

src/eva/multimodal/callbacks/writers/prediction.py
```diff
@@ -0,0 +1,39 @@
...
+            dataloader_idx_map=dataloader_idx_map,
+            metadata_keys=metadata_keys,
+            include_input=include_input,
+            overwrite=overwrite,
+            save_format=save_format,
+        )
+
+    @override
+    def _unpack_batch(self, batch: TextImageBatch) -> Tuple[list, list | None, dict | None]:  # type: ignore
+        text_batch, _, target_batch, metadata_batch = TextImageBatch(*batch)
```
MaxFeucht, line 38: At some point we might want to log images for expert validation of outputs in the context of the image and text input. Could we make the return of images an arg that defaults to False?

MaxFeucht (approved):
Great work, LGTM! Mostly minor comments, the image return might be important though

configs/language/pathology/offline/multiple_choice/pubmedqa.yaml
```diff
@@ -0,0 +1,70 @@
...
+  class_path: eva.Trainer
+  init_args:
+    n_runs: &N_RUNS ${oc.env:N_RUNS, 1}
+    default_root_dir: ${oc.env:OUTPUT_ROOT, logs/${oc.env:MODEL_NAME, anthropic-claude-3-7-sonnet-latest}/language/pubmedqa}
+    checkpoint_type: null
+    callbacks:
+      - class_path: eva.callbacks.ConfigurationLogger
+      - class_path: eva.language.callbacks.writers.TextPredictionWriter
+        init_args:
+          output_dir: &PREDICTIONS_OUTPUT_DIR ${oc.env:PREDICTIONS_OUTPUT_DIR, ./predictions/${oc.env:MODEL_NAME, anthropic-claude-3-7-sonnet-latest}/pubmedqa}
```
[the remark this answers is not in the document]
nkaenzig: Indeed, this could be done by setting `PREDICTIONS_OUTPUT_DIR` through the script that runs the evals

configs/language/pathology/offline/multiple_choice/pubmedqa.yaml
```diff
@@ -0,0 +1,70 @@
...
+          dataloader_idx_map:
+            0: val
```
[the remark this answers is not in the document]
nkaenzig: Good point, I just took this from the existing config, but didn't make sense indeed. I now listed val & test both explicitly.

src/eva/language/models/modules/language.py
```diff
@@ -53,3 +53,42 @@ def _batch_step(self, batch: TextBatch) -> STEP_OUTPUT:
...
+
+    def __init__(
+        self,
+        metrics: metrics_lib.MetricsSchema | None = None,
+        postprocess: batch_postprocess.BatchPostProcess | None = None,
+    ) -> None:
+        """Initializes the text inference module.
+
+        Args:
+            model: Model instance to use for forward pass.
```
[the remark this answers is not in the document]
nkaenzig: removed

src/eva/multimodal/callbacks/writers/prediction.py
MaxFeucht: Fair, I was thinking about a potential disconnect between data source and the logging location/service. But indeed, for internal use this should be possible and the way to go, and nice that this is supported already via metadata. If public users wish for such a functionality, we can still implement this at a later point. For now, let's leave it as is.

src/eva/language/data/datasets/prediction.py
MaxFeucht: Thanks for the explanation and the doc addition

PR #877, opened 2025-09-03
Fix wandb rename run issue with `wandb>=0.21.0`

### LINKED ISSUE #876: Run rename breaks logic with `wandb>=0.21.0`
Getting this error when calling the `rename_active_run` function:

```
TypeError: Run.save() missing 1 required positional argument: 'glob_str'
```

nkaenzig:
Closes #876

W&B now automatically syncs metadata periodically. In practice, assigning to wandb.run.name is enough in v0.21+ and an explicit `run.save()` call is not required anymore.

### DIFF
--- a/src/eva/core/loggers/utils/wandb.py
+++ b/src/eva/core/loggers/utils/wandb.py
@@ -5,6 +5,8 @@ from typing import Any, Dict
 
 from loguru import logger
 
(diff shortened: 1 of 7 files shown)


MaxFeucht (approved):
LGTM

PR #897, opened 2025-09-26
Ensure compatibility of transforms with `torchvision>=0.21.0`

### LINKED ISSUE #893: `torchvision.v2.transforms` breaking interface change in `>=0.21.0`
They've changed the interface of v2 transforms to be public, from _transform  to transform with this PR a while back: https://github.com/pytorch/vision/pull/8787.

This was released in either `0.21.0` or `0.22.0` (need to double check).

nkaenzig:
Closes #893

This PR adds a `TorchvisionTransformV2` base class which should be used for all transforms that follow the `torchvision.v2.transforms` API to ensure capability both with `>=0.21.0` and also older versions.


src/eva/vision/data/transforms/base/torchvision.py
```diff
@@ -0,0 +1,33 @@
...
+    as torchvision 0.21.0 introduced a new transform API where they
+    renamed the following methods:
+
+    - `get_params` -> `make_params`
+    - `_transform` -> `transform`
+    """
+
+    def make_params(self, flat_inputs: List[Any]) -> Dict[str, Any]:
+        """Called internally before calling transform() on each input."""
+        return {}
```
MaxFeucht, line 22: Is there are reason to return an empty dict here instead of an `NotImplementedError` like in `transform`?

src/eva/vision/data/transforms/base/torchvision.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class TorchvisionTransformV2(v2.Transform, abc.ABC):
+    """Wrapper for torchvision.v2.Transform.
+
+    This class ensures compatibility both with >=0.21.0 and older versions,
+    as torchvision 0.21.0 introduced a new transform API where they
+    renamed the following methods:
+
+    - `get_params` -> `make_params`
```
MaxFeucht, line 16: nit: `get_params` -> `_get_params`

MaxFeucht (approved):
LGTM! minor comments

PR #898, opened 2025-09-29
Make `BalancedSampler` compatible with str inputs for language and multimodal

MaxFeucht:
Makes the `BalancedSampler` compatible with `str` inputs by adding a conditional check

### DIFF
--- a/src/eva/core/data/samplers/classification/balanced.py
+++ b/src/eva/core/data/samplers/classification/balanced.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 from typing import Dict, Iterator, List, Union
 
 import numpy as np
+import torch
 from loguru import logger
 from typing_extensions import override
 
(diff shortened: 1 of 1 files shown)


src/eva/core/data/samplers/classification/balanced.py
```diff
@@ -72,10 +72,14 @@ def _make_indices(self):
                 _, target, _ = DataSample(*self.data_source[idx])
             if target is None:
                 raise ValueError("The dataset must return non-empty targets.")
-            if target.numel() != 1:
-                raise ValueError("The dataset must return a single & scalar target.")
-
-            class_idx = int(target.item())
+            if isinstance(target, str):
+                class_idx = target
+            elif hasattr(target, 'numel') and hasattr(target, 'item'):
```
nkaenzig, line 77: Maybe let's also specifically check if it's a tensor here:
`elif isinstance(target, torch.Tensor) and hasattr(target, 'numel') and hasattr(target, 'item')`

nkaenzig (approved):
LGTM

PR #931, opened 2025-10-30
Expand init args in `Resize` transform

nkaenzig:
- Expose `max_size`
- Allow both `max_bytes` and `size` to be specified, but if `max_bytes` is present it will take precedence

### DIFF
--- a/docs/user-guide/advanced/replicate_evaluations.md
+++ b/docs/user-guide/advanced/replicate_evaluations.md
@@ -7,7 +7,7 @@ You will need to download the config files and then in the following commands re
 
 Keep in mind:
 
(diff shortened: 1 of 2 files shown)


src/eva/vision/data/transforms/spatial/resize.py
```diff
@@ -22,34 +22,38 @@ class Resize(base.TorchvisionTransformV2):
...
 
         self.size = size
         self.max_bytes = max_bytes
+        self.max_size = max_size
         self.resize_fn = None
 
-        if size is not None:
-            self.resize_fn = v2.Resize(size=size)
-        elif max_bytes is not None:
+        if max_bytes is not None:
```
MaxFeucht: This automatically assumes precedence of the `max_bytes`, we could add a check for if both are given, such as 


```
if (size is not None and max_bytes is not None) or (
            max_size is not None and max_bytes is not None
        ):
            raise ValueError(
                "Cannot provide spatial size parameters 'size'/'max_size'"
                " and 'max_bytes' at the same time."
            )
```

PR #932, opened 2025-10-30
Unify prompting and answer extraction logic and introduce factories

MaxFeucht:
This PR unifies the prompting templates for multiple choice and freeform QA for all three currently supported answer formats: `JSON`, `XML`, and `raw` (a largely unstructured output format).

For each template, one correspnding extraction class is added. For both the template and extraction classes, a factory logic is added that serves the respective classes based on an `answer_format` argument, to allow for dynamic switching between formats within a config via env vars.

## Changes:
- Adds common abstraction for extraction classes for freeform and discrete (MC) answers
- Adds new prompting templates such that all three formats are supported in MCQA and FFQA
- Adds a parsing function for unconstrained (`raw`) and `xml` outputs
- Adds factories for both prompting templates and extraction classes to dynamically switch between formats
- Adapts the `apply_transforms` function in post processing to accommodate factories
- Adapts existing Multiple Choice configs to use new factory setup

## Notes:
- The current parsing approach for MCQA tasks is only stable for when using `use_option_letters = True`. This is due to the static mapping that is defined within the configs for MCQA tasks, as well as due to the conception of the `extract_raw` function for the `raw` formatting variant.

### DIFF
--- a/src/eva/language/models/postprocess/extract_answer/factory.py
+++ b/src/eva/language/models/postprocess/extract_answer/factory.py
@@ -7,7 +7,10 @@ from eva.language.models.postprocess.extract_answer.json import (
     ExtractAnswerFromJson,
     ExtractDiscreteAnswerFromJson,
 )
-from eva.language.models.postprocess.extract_answer.raw import ExtractDiscreteAnswerFromRaw
+from eva.language.models.postprocess.extract_answer.raw import (
+    ExtractAnswerFromRaw,
+    ExtractDiscreteAnswerFromRaw,
+)
 from eva.language.models.postprocess.extract_answer.xml import (
     ExtractAnswerFromXml,
     ExtractDiscreteAnswerFromXml,
@@ -60,5 +63,7 @@ class ExtractAnswer:
                 return ExtractAnswerFromJson(**extract_kwargs)
             case "xml":
                 return ExtractAnswerFromXml(**extract_kwargs)
+            case "raw":
+                return ExtractAnswerFromRaw(**extract_kwargs)
             case _:
                 raise ValueError(f"Unknown answer format: {answer_format}")
--- a/src/eva/language/models/postprocess/extract_answer/raw/__init__.py
+++ b/src/eva/language/models/postprocess/extract_answer/raw/__init__.py
@@ -1,7 +1,9 @@
 """API for extracting answers from raw model outputs."""
 
 from eva.language.models.postprocess.extract_answer.raw.discrete import ExtractDiscreteAnswerFromRaw
+from eva.language.models.postprocess.extract_answer.raw.free_form import ExtractAnswerFromRaw
 
 __all__ = [
     "ExtractDiscreteAnswerFromRaw",
+    "ExtractAnswerFromRaw",
 ]
--- a/src/eva/language/models/postprocess/extract_answer/raw/free_form.py
+++ b/src/eva/language/models/postprocess/extract_answer/raw/free_form.py
@@ -1,0 +1,21 @@
+"""Postprocessing transforms for extracting answers from raw responses."""
+
+from typing_extensions import override
+
+from eva.language.models.postprocess.extract_answer.base import ExtractAnswerFromStructuredOutput
+
+
+class ExtractAnswerFromRaw(ExtractAnswerFromStructuredOutput):
+    """Extracts answers from raw responses and returns structured data."""
+
+    @override
+    def _extract_structured_data(self, value: str) -> str:
+        """Extract raw data from a string; returning the string as is.
+
+        Args:
+            value: The input string.
+
+        Returns:
+            str: The input string unchanged.
+        """
+        return value
--- a/tests/eva/language/models/postprocess/extract_answer/raw/test_free_form.py
+++ b/tests/eva/language/models/postprocess/extract_answer/raw/test_free_form.py
@@ -1,0 +1,28 @@
+"""Tests for ExtractAnswerFromRaw post-processing transform."""
+
+from eva.language.models.postprocess.extract_answer.raw.free_form import ExtractAnswerFromRaw
+
+
+def test_single_string_returns_wrapped_in_list() -> None:
+    """Single string input should be wrapped in a list unchanged."""
+    transform = ExtractAnswerFromRaw()
+    result = transform("Test response")
+
+    assert result == ["Test response"]
+
+
+def test_list_input_returns_unchanged() -> None:
+    """List of strings should be returned unchanged."""
+    transform = ExtractAnswerFromRaw()
+    input_list = ["First", "Second", "Third"]
+    result = transform(input_list)
+
+    assert result == input_list
+
+
+def test_accepts_optional_parameters() -> None:
+    """Should accept optional parameters for compatibility."""
+    transform = ExtractAnswerFromRaw(answer_key="custom", case_sensitive=False)
+    result = transform("Test")
+
+    assert result == ["Test"]


src/eva/language/utils/text/raw.py
```diff
@@ -0,0 +1,95 @@
...
+
+    Focuses on the last few lines of the response where models typically place their final answer.
+
+    Args:
+        value: The input string containing the model's response.
+        options: Optional list of valid multiple-choice options (e.g., ["A", "B", "C"]).
+                If None, will attempt to extract any single letter options.
+
+    Returns:
+        Dict[str, str] | None: Dictionary with 'answer' key containing the extracted option,
```
nkaenzig: let's make the key configurable (`answer_key`) and use `answer` as default

src/eva/language/utils/text/raw.py
```diff
@@ -0,0 +1,95 @@
...
+        text: Cleaned text to extract answer from.
+        options: List of valid options. If None, defaults to A-Z.
+
+    Returns:
+        The extracted option or None if no valid answer found.
+    """
+    if not text:
+        return None
+
+    if options is None:
```
nkaenzig: I think we should make the `options` argument mandatory (no default), then we can remove this if and A-Z fallback.

src/eva/language/utils/text/raw.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # Distinguish between single-character and multi-character options
+    all_single_char = all(len(opt) == 1 for opt in options)
+    option_regex = (
+        f"([{''.join(re.escape(opt.lower()) for opt in options)}])"
+        if all_single_char
+        else f"({'|'.join(re.escape(opt.lower()) for opt in options)})"
+    )
+
+    patterns = [
```
nkaenzig: maybe we can simplify those into one pattern, focusing on whether the answer occurs somewhere in an "isolated" fashion in the tail.

chatgpt suggests this:

```python
# Build alternation: longer options first to avoid prefix capturing.
alternation = "|".join(re.escape(o) for o in sorted_opts)

# "Isolated": not preceded/followed by a word char. This works for both single letters
# and multi-word options (e.g., "Option 2"). It allows punctuation/whitespace around.
pattern = re.compile(rf"(?<!\w)({alternation})(?!\w)", flags=re.IGNORECASE)
```

src/eva/language/utils/text/raw.py
```diff
@@ -0,0 +1,95 @@
...
+            return {"answer": extracted_answer.upper()}
+
+        return None
+
+    except Exception as e:
+        logger.warning(f"Failed to extract answer from text: {e}")
+        return None
+
+
+def _get_text_tail(text: str, max_chars: int = 100) -> str:
```
nkaenzig: I'd consider to instead of using number of characters `max_chars`, number of words to specify the tail length. Also, I wouldn't set this too high, maybe to 5 words max.

tests/eva/language/models/postprocess/extract_answer/discrete/test_discrete_answer_extraction_from_json.py.py
nkaenzig: plz make sure to mirror the module / folder structure from the lib code in the unit tests. Also the naming convention of the test files is always `test_<name_of_the_lib_file>.py`

nkaenzig (commented):
Looks very nice, I left a couple of comments.

src/eva/core/models/modules/utils/batch_postprocess.py
```diff
@@ -56,7 +56,13 @@ def _apply_transforms(tensor: torch.Tensor, transforms: List[Transform]) -> torc
     Returns:
         The processed tensor.
     """
-    return functools.reduce(lambda tensor, transform: transform(tensor), transforms, tensor)
+
+    def apply_transform(tensor: torch.Tensor, transform: Transform) -> torch.Tensor:
```
nkaenzig: why is this needed / i.e. when can this happen?

src/eva/language/utils/text/raw.py
```diff
@@ -0,0 +1,95 @@
...
+    if not text:
+        return None
+
+    if options is None:
+        options = [chr(i) for i in range(ord("A"), ord("Z") + 1)]
+
+    # Distinguish between single-character and multi-character options
+    all_single_char = all(len(opt) == 1 for opt in options)
+    option_regex = (
+        f"([{''.join(re.escape(opt.lower()) for opt in options)}])"
```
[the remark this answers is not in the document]
MaxFeucht: The other extraction classes are not case sensitive, since the parsing doesn't depend on the parsed content, but I agree that it makes sense to add it in the regex case. The `answer_key` is also automatically determined for XML and JSON based on the parsed keys

src/eva/language/utils/text/raw.py
MaxFeucht: The fallback might not be needed actually, since we pass the mapping keys as options in the `ExtractDiscreteAnswerFromRaw` class by default

src/eva/language/utils/text/raw.py
MaxFeucht: Good idea, implemented this now. I set the default to 10 for the case when the response might be given in the form of `"A) Carcinoma is a commonly occurring disease"`, i.e., when the option text is outputted as well and might be long in some cases

src/eva/language/utils/text/raw.py
```diff
@@ -0,0 +1,95 @@
+"""Raw text extraction utilities."""
+
+import re
+
+from loguru import logger
+
+
+def extract_raw(value: str, options: list[str] | None = None) -> dict | None:
```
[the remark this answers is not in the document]
MaxFeucht: Renamed

src/eva/language/prompts/templates/json/multiple_choice.py
```diff
@@ -20,15 +21,14 @@ class JsonMultipleChoicePromptTemplate(base.PromptTemplate):
         """\
         {{ preamble }}
 
+
```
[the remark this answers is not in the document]
MaxFeucht: Adapted, good catch

src/eva/language/prompts/templates/json/multiple_choice.py
```diff
@@ -40,11 +40,23 @@ class JsonMultipleChoicePromptTemplate(base.PromptTemplate):
...
+        ---
+        {% endfor %}
+        Now please answer the initial question.
+        {% else %}
         Example JSON Answer:
+        Your explanation for why you chose this answer can go here...
         {{ '{' }}
-            "{{ answer_key }}": "{{ example_answer }}",
-            "{{ reason_key }}": "{{ example_reason }}"
+            "{{ answer_key }}": "Your answer here"
```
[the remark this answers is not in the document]
MaxFeucht: Added `example_answer` back in and for all templates

src/eva/language/prompts/templates/xml/free_form.py
```diff
@@ -0,0 +1,101 @@
...
+        {% if examples %}
+        Below are some examples:
+
+        {% for ex in examples %}
+        Example {{ loop.index }}:
+        Question: {{ ex.question }}
+        Answer: {{ ex.answer }}
+        ---
+        {% endfor %}
+        Now please answer the initial question.
```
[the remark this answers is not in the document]
nkaenzig: Amazing, I think the templates look great now.

src/eva/language/utils/text/raw.py
nkaenzig: My point is that we should never fallback to A-Z options. If it's a multiple choice task that uses option letters, we can set that explicitly through init args.

src/eva/language/utils/text/raw.py
```diff
@@ -0,0 +1,138 @@
...
+
+    try:
+        text = text.strip()
+        tail_text = _get_text_tail(text, max_words=tail_length)
+        cleaned_text = _clean_string(tail_text)
+
+        if answer_options is not None:
+            options = answer_options
+        else:
+            options = [chr(i) for i in range(ord("A"), ord("Z") + 1)]
```
nkaenzig: Same comment as elsewhere, I think we should never use fallback to A-Z, but instead set explicitly through init args whether it's a task that uses option letters.

configs/language/pathology/online/multiple_choice/pubmedqa.yaml
```diff
@@ -27,9 +27,11 @@ model:
       predictions_transforms:
         - class_path: eva.language.models.postprocess.ExtractDiscreteAnswerFromJson
```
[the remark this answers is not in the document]
nkaenzig: I guess the issue is that the extract classes now return a dict, while `MulticlassClassificationMetrics` expects a tensor.

src/eva/language/models/postprocess/extract_answer/base.py
```diff
@@ -0,0 +1,144 @@
...
+    def __call__(
+        self, values: Union[str, List[str]]
+    ) -> Dict[str, Union[torch.Tensor, List[Dict[str, str] | None]]]:
+        """Convert structured string(s) to a tensor of integer labels."""
+        structured_data = super().__call__(values)
+        answers = list(map(self._extract_answer, structured_data))
+
+        return {
+            "predictions": torch.tensor(answers, dtype=torch.long),
+            "extracted_answers": structured_data,
```
nkaenzig: why returning a dict here? leads to the issue here: https://github.com/kaiko-ai/eva/pull/932/files#r2485662139

src/eva/language/models/postprocess/extract_answer/json/free_form.py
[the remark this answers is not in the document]
MaxFeucht: Adapted this

PR #948, opened 2025-11-13
Fix/litellm text only inputs

### LINKED ISSUE #947: LiteLLM wrapper fails with TypeError on text-only inputs
### Description

The LiteLLM model wrapper crashes when processing text-only datasets (e.g., HealthBench) due to improper handling of None image batches:

```
  Error:
  TypeError: 'NoneType' object is not iterable
    File "eva/multimodal/models/wrappers/litellm.py", line 59, in format_inputs
      map(..., message_batch, image_batch)
```

When processing text-only inputs, unpack_batch() returns image_batch=None. The format_inputs() method attempts to iterate over this None value in map(), causing the TypeError.

Solution:

  Convert None to a list of None values before calling map():

```
  # litellm.py lines 57-59
  if image_batch is None:
      image_batch = [None] * len(message_batch)
```
  This allows map() to iterate properly, and format_litellm_message() already handles None images correctly by formatting text-only messages.

Impact:
  Fixes 8 text-only benchmarks (HealthBench, PubMedQA, etc.) for all API-based models (OpenAI, Anthropic, Google). evaluation configurations now work that might previously failed.

sebastianffx:
Fixes #947


src/eva/multimodal/models/wrappers/litellm.py
```diff
@@ -55,6 +55,10 @@ def format_inputs(self, batch: TextImageBatch) -> List[List[Dict[str, Any]]]:
         )
         message_batch = list(map(language_message_utils.combine_system_messages, message_batch))
 
+        # Handle text-only inputs where image_batch is None
```
nkaenzig: let's remove the inline comment here

PR #953, opened 2025-11-20
Enable logging of input text for HF and LiteLLM models

MaxFeucht:
Enables logging of input text for Huggingface and LiteLLM models by adding an optional `input_text` field to the `ModelOutput` type. For the `HuggingFaceModel` this field is filled with the decoded input **including** special tokens, for the LiteLLM model we simply concatenate the role and contents of the single messages for logging.

Changes:
- Adapted `ModelOutput` type
- Added `stringify_messages` as a more general case of `merge_message_contents` that also handle messages in dict format and includes optional role addition
- Added `input_text` to `multimodal.HuggingFaceModel` and `language.LiteLLMModel`

### DIFF
--- a/src/eva/multimodal/models/wrappers/huggingface.py
+++ b/src/eva/multimodal/models/wrappers/huggingface.py
@@ -144,7 +144,6 @@ class HuggingFaceModel(base.VisionLanguageModel):
         return ModelOutput(
             generated_text=decoded_output,
             input_text=decoded_input,
-            input_ids=batch.get("input_ids"),
             output_ids=output_ids,
             attention_mask=batch.get("attention_mask"),
         )


src/eva/language/utils/text/messages.py
```diff
@@ -39,21 +39,35 @@ def combine_system_messages(message: MessageSeries, join_char: str = "\n") -> Me
...
     non_system_messages = list(filter(lambda item: item.role != Role.SYSTEM, message))
     return [
-        SystemMessage(content=merge_message_contents(system_messages, join_char=join_char))
+        SystemMessage(content=merge_messages(system_messages, join_char=join_char))
     ] + non_system_messages
 
 
-def merge_message_contents(message: MessageSeries, join_char: str = "\n") -> str:
-    """Merges the all contents within a message series into a string.
+def merge_messages(
```
nkaenzig: This function was mainly written for the purpose where we have multiple system messages that need to be combined into one, i.e. merging message contents without altering the content as such.

For logging purposes, we could also use the existing `serialize` function, which includes the roles - or we could have a dedicated function that "pretty" prints message series.

MaxFeucht: Yeah I see, I was not quite sure what would be the best way to go here - I decided to collapse multiple use-cases into one function since the logic was very similar. I can try to preserve `merge_message_content` and use the `serialize` function, and if that doesn't do it create a logging-specific message merging func.

src/eva/multimodal/models/wrappers/huggingface.py
```diff
@@ -139,9 +139,11 @@ def model_forward(self, batch: Dict[str, torch.Tensor]) -> ModelOutput:
             A dictionary containing the processed input and the model's output.
         """
         output_ids = self.model.generate(**batch, **self.generation_kwargs)  # type: ignore
+        decoded_input, decoded_output = self._decode_ids(output_ids, batch["input_ids"].shape[-1])
```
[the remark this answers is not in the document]
nkaenzig: Yes that's the convention indeed, but in general the model forward pass implementation is out of our control and people could potentially violate this. 

Given that we already log `batch["input_ids"]` in the `input_ids` field of the model output, it might be more consistent to decode those ids - this way we are 100% sure that `input_ids` and `input_text` refer to the same, and that this was what we effectively sent into the model.

src/eva/language/utils/text/messages.py
MaxFeucht: Separated the two now into `stringify_messages` and the more specialized case of `merge_message_contents` as it was before

PR #973, opened 2025-12-15
added boxed formatting

jonaserb-k:
## Summary

Adds support for the `boxed` answer format, which extracts answers wrapped in `\boxed{}` notation. This format is used by a few of the models, especially [OctoMed](https://huggingface.co/OctoMed/OctoMed-7B).

## Changes

- **Answer Extraction**: Added `ExtractAnswerFromBoxed` (free-form) and `ExtractDiscreteAnswerFromBoxed` (MCQA) classes
- **Templates**: Added `boxed` templates for both multiple-choice and free-form questions
- **Factory Support**: Registered `boxed` format in the answer extraction factory
- **Tests**: Added 51 comprehensive tests covering extraction logic and template rendering

## Key Features

- Extracts content from `\boxed{answer}` notation
- Supports markdown code fences (`latex`, `math`, or plain)
- Handles whitespace and surrounding text
- Uses the last \boxed{} entry if there are multiple in the answer string
- Handles nested {} braces such that entries like \boxed{\frac{1}{2}} answers are handled correctly

## Testing

- All 51 tests passing ✅
- ran the formatting in multiple test runs with correct output

```bash
pdm run pytest tests/eva/language/models/postprocess/extract_answer/boxed/ tests/eva/language/prompts/templates/boxed/ -v

### DIFF
--- a/src/eva/language/models/constants.py
+++ b/src/eva/language/models/constants.py
@@ -2,5 +2,5 @@
 
 import os
 
-MAX_NEW_TOKENS = int(os.getenv("MAX_NEW_TOKENS", 8192))
+MAX_NEW_TOKENS = int(os.getenv("MAX_NEW_TOKENS", 32768))
 """Default maximum number of new tokens to generate for language models."""
--- a/src/eva/language/models/modules/language.py
+++ b/src/eva/language/models/modules/language.py
@@ -34,7 +34,7 @@ class LanguageModule(module.ModelModule):
 
     @override
     def configure_model(self) -> None:
-        model = self.model.model if hasattr(self.model, "model") else self.model
+        model = self.model.model if type(self.model).__name__ == "ModelFromRegistry" else self.model
         if hasattr(model, "configure_model"):
             model.configure_model()  # type: ignore
 
--- a/src/eva/language/models/networks/__init__.py
+++ b/src/eva/language/models/networks/__init__.py
@@ -6,6 +6,8 @@ from eva.language.models.networks.api import (
     Claude37Sonnet20250219,
     Gemini25FlashLite,
 )
+from eva.language.models.networks.meta import Lllama3_2_3BInstruct
+from eva.language.models.networks.microsoft import Phi3Mini4KInstruct
 from eva.language.models.networks.registry import model_registry
 
 __all__ = [
@@ -13,5 +15,7 @@ __all__ = [
     "Claude37Sonnet20250219",
     "Gemini25FlashLite",
     "Qwen205BInstruct",
+    "Phi3Mini4KInstruct",
+    "Lllama3_2_3BInstruct",
     "model_registry",
 ]
--- a/src/eva/language/models/networks/alibaba.py
+++ b/src/eva/language/models/networks/alibaba.py
@@ -20,6 +20,7 @@ class Qwen205BInstruct(wrappers.HuggingFaceModel):
         requirements.check_min_versions(requirements={"torch": "2.5.1", "torchvision": "0.20.1"})
         super().__init__(
             model_name_or_path="Qwen/Qwen2-0.5B-Instruct",
+            model_class="AutoModelForCausalLM",
             model_kwargs={
                 "torch_dtype": torch.bfloat16,
             },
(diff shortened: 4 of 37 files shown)

### CODE REVIEW

File: src/eva/language/models/postprocess/extract_answer/factory.py [hunk 1]
```diff
@@ -21,12 +25,12 @@ class ExtractDiscreteAnswer:
     """Factory for creating discrete answer extractors."""
 
     def __new__(
-        cls, answer_format: Literal["json", "xml", "raw"], extract_kwargs: dict
+        cls, answer_format: Literal["json", "xml", "raw", "boxed"], extract_kwargs: dict
```
MaxFeucht: optional nit: put `raw` last since it's the least structured format

src/eva/language/utils/text/boxed.py
```diff
@@ -0,0 +1,75 @@
...
+        if text[i] == "{":
+            count += 1
+        elif text[i] == "}":
+            count -= 1
+        i += 1
+    return i - 1 if count == 0 else -1
+
+
+def extract_boxed(response: str, raise_if_missing: bool = False) -> Dict[str, str] | None:
+    r"""Extracts content from \\boxed{} tags and converts to a dictionary.
```
MaxFeucht: Why return a dict by default if the boxed format doesn't dictate any structure of the answer?

src/eva/language/utils/text/boxed.py
```diff
@@ -0,0 +1,75 @@
...
+        else:
+            clean_response = response.strip()
+
+        # Find last \boxed{
+        boxed_start = clean_response.rfind("\\boxed{")
+        if boxed_start == -1:
+            raise ValueError("No \\boxed{} content found.")
+
+        # Find the content between the braces
+        content_start = boxed_start + 7  # len('\\boxed{')
```
MaxFeucht: Can there be cases when a model only outputs `\boxed{`? Would be quite relevant for single letter answers

jonaserb-k: Good point. I was relying here on [this](https://github.com/PrimeIntellect-ai/verifiers/blob/ac2b2e95e7668f184e497524e546900fffca6bae/verifiers/utils/data_utils.py#L72) tried and tested boxed implementation of PrimeIntellect-AI.

The +7 is correct if we match the exact marker \boxed{ (which we currently do via `rfind("\\boxed{")`). To make it future-proof/cleaner I could replace the constant with len(marker) (e.g. `marker="\\boxed{"; content_start=boxed_start+len(marker)`). This would make sense if we intended to use variations if the models outputs an incorrect marker like `boxed\{` or `\\boxed` (printing two literal backslashes). Though then the nesting logic would also have to be adapted to the adapted markers and the question is should we really accept other "wrong" variants of the "boxed{" marker?

src/eva/language/utils/text/boxed.py
MaxFeucht: Hmmm good point, actually the typing here makes only sense for the discrete case (in which we extract then the value from the returned dict, [here](https://github.com/kaiko-ai/eva/blob/main/src/eva/language/models/postprocess/extract_answer/base.py#L125)). For the freeform case we currently return the extracted output as is, so we end up with `"{"answer": freeform response}"` as model output for `json`/`xml` and now `boxed`. The `raw` formatting also doesn't adhere to the Dict typing, see [here](https://github.com/kaiko-ai/eva/blob/main/src/eva/language/models/postprocess/extract_answer/raw/free_form.py). 

We could enforce dict returns for all extraction functions and add an explicit value extraction in the base class, by moving the `_extract_answer` method from `ExtractDiscreteAnswerFromStructuredOutput` to `ExtractAnswerFromStructuredOutput`. What do you think @jonaserb-k and @nkaenzig?

src/eva/language/utils/text/boxed.py
```diff
@@ -0,0 +1,75 @@
...
+        closing_brace = _find_matching_brace(clean_response, content_start)
+
+        if closing_brace == -1:
+            raise ValueError("No matching closing brace found.")
+
+        answer = clean_response[content_start:closing_brace].strip()
+        boxed_dict = {"answer": answer}
```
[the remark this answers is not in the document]
MaxFeucht: > Absolutely right, implemented the answer key to be configurable. I removed configurability 

I'm not sure I'm following 😅 Let's continue the answer key and dict return discussion in the [other comment](https://github.com/kaiko-ai/eva/pull/973#discussion_r2627499700)

src/eva/language/utils/text/boxed.py
jonaserb-k: as discussed this fix will be implemented after the merge of this PR since it is not "boxed" specific

src/eva/language/utils/text/boxed.py
jonaserb-k: reference added.

PR #977, opened 2025-12-18
Fix inconsistency in implementation of extract classes

### LINKED ISSUE #976: Inconsistency in implementation of `Extract` classes
See https://github.com/kaiko-ai/eva/pull/973#discussion_r2623605616 for context.

nkaenzig:
Closes #976

## What this PR does?
- Make sure that all implementations of the abstract function `_extract_structured_data` of `ExtractAnswerFromStructuredOutput` return a dictionary
- By default, now `ExtractAnswerFromStructuredOutput` only returns the string contained in the `self.answer_key` entry within that dictionary. There is also a `return_dict` arg now, which when set to `True` will return the full dictionary (this is the same as the previous behaviour, and might be helpful for analytics / logging purposes, but in my opinion should not be the default, as for the metrics calculation only the final answer should be considered)

### DIFF
--- a/src/eva/language/models/postprocess/extract_answer/base.py
+++ b/src/eva/language/models/postprocess/extract_answer/base.py
@@ -87,7 +87,7 @@ class ExtractAnswerFromStructuredOutput(ABC):
             )
             return self.missing_answer
 
-        return structured_data if self.return_dict else structured_data[self.answer_key]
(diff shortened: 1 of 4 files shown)


MaxFeucht (approved):
LGTM! Thanks for taking care of that

PR #996, opened 2026-03-11
Add record_datasets_as_runs setting to trainer

### LINKED ISSUE #995: Evals where multiple prediction datasets are not supported
Offline evals like `configs/language/pathology/offline/multiple_choice/pubmedqa.yaml` where predictions are pre-generated and then loaded via a prediction dataset, currently only support a single prediction dataset.

However, it might make sense to generate multiple sets of predictions with different seeds / temperature parameters, and then have eva calculate the metrics individually for these but also calculating mean/std aggregations.

nkaenzig:
## Summary
- Adds a `record_datasets_as_runs` parameter to `Trainer` for evaluation sessions with multiple `val` and/or `test` datasets.
- When enabled, each dataset output is recorded as a separate run in the session summary instead of being kept as separate datasets within a single run.
- Passes the setting through the `Interface` CLI entry points and `run_evaluation_session`.

## How to use

Set `record_datasets_as_runs: true` in the trainer config, and provide a list of datasets in the `val` or `test` section of the data config. Each dataset entry can point to a different data path, for example predictions from different offline runs:

```yaml
trainer:
  class_path: eva.Trainer
  init_args:
    record_datasets_as_runs: true
    # ... other trainer args

data:
  class_path: eva.DataModule
  init_args:
    datasets:
      val:
        - class_path: eva.language.datasets.TextPredictionDataset
          init_args:
            path: ./predictions/run_1/manifest.jsonl
            split: val
        - class_path: eva.language.datasets.TextPredictionDataset
          init_args:
            path: ./predictions/run_2/manifest.jsonl
            split: val
        - class_path: eva.language.datasets.TextPredictionDataset
          init_args:
            path: ./predictions/run_3/manifest.jsonl
            split: val
```

In the generated `results.json`, metrics are reported as one dataset entry whose `values` list contains the per-dataset run values together with the aggregated `mean` and `stdev`:

```json
{
  "metrics": {
    "val": [
      {
        "val/AverageLoss": {
          "mean": 1.58,
          "stdev": 0.0,
          "values": [1.60, 1.55, 1.59]
        },
        "val/BinaryBalancedAccuracy": {
          "mean": 0.50,
          "stdev": 0.01,
          "values": [0.49, 0.51, 0.50]
        }
      }
    ]
  }
}
```


DriessenA:
@nkaenzig Any chance we can merge this PR? I'd be happy to help if needed :)

MaxFeucht (approved):
LGTM

PR #1004, opened 2026-03-17
Optimize NiFTi file read and add unit-tests

ioangatop:
Closes #1001


src/eva/vision/utils/io/nifti.py
```diff
@@ -167,3 +223,37 @@ def _reorient(
...
+            - UniformSampler: Returns n indices spaced evenly across the depth.
+            - GaussianSampler: Returns n indices according to a Gaussian distribution.
+
+    Returns:
+        A list of integer slice indices to extract.
+    """
+    total = data_shape[-1]
+    match sampler:
+        case IndexSampler(indices):
+            return indices
```
nkaenzig: `indices` doesn't seem to be defined at this point?

src/eva/vision/utils/io/nifti.py
```diff
@@ -167,3 +223,37 @@ def _reorient(
...
+        case BlockSampler(n):
```
nkaenzig: `n` doesn't seem to be defined at this point?

src/eva/vision/utils/io/nifti.py
```diff
@@ -41,8 +116,17 @@ def read_nifti(
...
     _utils.check_file(path)
     image_data = _load_nifti_silently(path)
-    if slice_index is not None:
-        image_data = image_data.slicer[:, :, slice_index : slice_index + 1]
+
+    if sampler:
+        total = image_data.shape[-1]
+        slice_indices = sampler.sample(total)
+        proxy_slices = [image_data.dataobj[:, :, i] for i in slice_indices]
+        image_data = nib.nifti1.Nifti1Image(
```
nkaenzig: The sampler path here constructs a new image from sampled slices but reuses the original affine/header unchanged. For anything except the untouched full volume, that spatial metadata is no longer valid, so downstream consumers will interpret coordinates incorrectly.

nkaenzig (approved):

PR #1010, opened 2026-03-27
[Radiology] Add ct scan level classification datasets

### LINKED ISSUE #1009: [Radiology] Add ct-scan level classification datasets
Add datasets `OrganMNIST3D`, `NoduleMNIST3D`, `LUNA25` and `ChestCTCancer` datasets

ioangatop:
Closes #1009 

Tests for mnist datasets omitted, as they are just wrappers, luna25 can not use the data for testing

Note: Luna25 datasets needs registration so no automatic downloads


pyproject.toml
```diff
@@ -73,6 +73,8 @@ vision = [
     "scipy>=1.14.0",
     "monai>=1.3.2",
     "einops>=0.8.1",
+    "medmnist>=3.0.2",
+    "scikit-learn>=1.7.2",
```
nkaenzig: sklearn might be a heavy dependency to add here, especially if we only need it for this stratified splitting logic. Note also that eva has a `splitting` module: src/eva/core/data/splitting/stratified.py - would it be an option to add group based splitting capabilities there instead?

ioangatop: nice, done!

src/eva/core/data/splitting/stratified.py
```diff
@@ -21,16 +22,42 @@ def stratified_split(
...
     if train_ratio + val_ratio + (test_ratio or 0) > 1.0:
         raise ValueError("The sum of the ratios must be lower or equal to 1.")
 
+    if groups is not None:
+        unique_groups, group_indices = np.unique(groups, return_inverse=True)
+        group_targets = np.array(
+            [targets[np.where(group_indices == i)[0][0]] for i in range(len(unique_groups))]
+        )
+
+        train_g, val_g, test_g = stratified_split(
```
nkaenzig, line 50: Why using recursion here?

src/eva/core/data/splitting/stratified.py
```diff
@@ -11,6 +11,7 @@ def stratified_split(
     train_ratio: float,
     val_ratio: float,
     test_ratio: float = 0.0,
+    groups: Sequence[Any] | None = None,
```
nkaenzig: Can you please expand to unit tests with test cases for this?

src/eva/core/data/splitting/stratified.py
ioangatop: recursion is used to reuse the same stratified splitting logic at the group level, ensuring consistent stratification without duplicating code for a second “group-aware” implementation.
This allows us to first split groups as atomic units, then map the result back to sample indices so that all samples within a group remain in the same split.

note that we alter the `targets=group_targets` and `groups=None`, so the flow of the function always goes to the second part of the function

src/eva/core/data/splitting/stratified.py
ioangatop: good point, done

pyproject.toml
```diff
@@ -102,6 +103,8 @@ all = [
     "litellm>=1.61.8",
     "backoff>=2.2.1",
     "json-repair>=0.52.0",
+    "medmnist>=3.0.2",
+    "scikit-learn>=1.7.2",
```
nkaenzig: sklearn dependency is still here? (https://github.com/kaiko-ai/eva/pull/1010#discussion_r3043606338)

nkaenzig: Also might need to regenerate the lock file after removing it.

nkaenzig (approved):

pyproject.toml
ioangatop: yes I believe I did it already but I'll check!

PR #1019, opened 2026-04-28
fix: Ensure `max_size` acts as upper bound

MaxFeucht:
In `eva`'s `Resize` implementation, the `max_size` arg was intended to act as an upper bound. When `size` is `None`, however, `max_size` acts as a `size` param causing smaller images to be upscaled. This PR ensures `max_size` when provided in isolation indeed acts as an upper bound.


src/eva/vision/data/transforms/spatial/resize.py
```diff
@@ -79,5 +81,16 @@ def _(self, inpt: Any, params: Dict[str, Any]) -> Any:
         if not self.resize_fns:
             return inpt
         for resize_fn in self.resize_fns:
+            if self._skip_max_size(resize_fn, inpt):
+                continue
             inpt = resize_fn(inpt)
         return tv_tensors.wrap(inpt, like=inpt)
+
+    def _skip_max_size(self, resize_fn: Any, inpt: Any) -> bool:
```
nkaenzig: nit: woul rename to `_skip_resize`

nkaenzig (approved):
LGTM

src/eva/vision/data/transforms/spatial/resize.py
MaxFeucht: Done

